PR #32, opened 2026-07-28
autoimprove: quarantine reconciliation + clear stale Ethiopia rows

### LINKED ISSUE #20: Clear stale quarantine entries (Ethiopia) and add quarantine reconciliation
Two assertions sit in `pipelines/polity-autoimprove/state/quarantine.csv` from before the Ethiopia coverage gap was closed:

```
ethiopia|fao1952|1937-1937   candidate AOI-1936-1941
ethiopia|iia|1938-1938       candidate AOI-1936-1941
```

They were quarantined because two agents agreed on the *substance* — that `iia` and `fao1952` both keep Ethiopia, Eritrea and Somaliland as separate labels, so "Ethiopia" means Ethiopia proper, not the Italian East Africa aggregate — but disagreed on the remedy (one proposed rerouting to an existing ETH row, the other proposed a new polity).

**The underlying gap has since been closed**: `ETH-1936-1941` was created, and both labels now route there (verified). So these rows are stale quarantine entries, not open questions.

**Work:** confirm both keys now resolve to `ETH-1936-1941` and clear them from `quarantine.csv`, or re-verify them under the current polity family so they bank properly.

More generally, `apply_verdicts.py` writes to `quarantine.csv` but nothing ever clears it — resolved entries accumulate. Worth adding a reconciliation step that drops quarantine rows whose assertion now banks cleanly, so the file stays an actionable queue rather than a log.

lbm364dl:
## What

`apply_verdicts.py` writes rows to `state/quarantine.csv` (two agents disagreed, or a verdict failed an execution contract) but nothing ever cleared them, so rows outlived the situations they recorded and the file stopped being an actionable queue. This adds the clearing pass and runs it.

## Where the logic lives, and why

New standalone script: **`pipelines/polity-autoimprove/reconcile_quarantine.py`** — not a `--reconcile` flag on `apply_verdicts.py`.

Reasons:

- `apply_verdicts.py` is *"apply one verdicts file"*. It hard-loads `verdicts_pending*.json` **and** `state/assertions.json` at module top level; both are gitignored per-run artefacts. A `--reconcile` flag would have to short-circuit around its own required inputs, which is exactly the shape that rots.
- Reconciliation must be runnable **when no verification run is in flight** — which is the normal case for housekeeping (and was the case here: there is no `assertions.json` in the tree right now). The new script therefore treats `assertions.json` as optional and falls back to re-deriving the route through `matchlib` (the same rule set `00_intake.py` uses) from the polities DB + `applied_aliases.csv`.
- Different responsibility and different blast radius: `apply_verdicts.py` only ever *adds* state; this pass *removes* rows, and deserves to be invoked deliberately (it also has `--dry-run`).

It mirrors `apply_verdicts.py`'s conventions: same `state/` path block, `TODAY` constant, and an `append_dedup` helper (extended with a `dedup_on` subset so the audit trail ignores `resolved_date`).

## Resolution rules

A quarantine row is dropped when either:

- **`banked`** — `state/review_ledger.csv` now shows the assertion `correct`/`fixed`: a later verification banked it, nothing left to adjudicate.
- **`route_changed`** — the assertion no longer routes to the `candidate` recorded in the row. The disagreement was *about that candidate*; once a new polity, an applied alias or a matcher fix sends the label elsewhere, the recorded dispute is about a routing that no longer exists, and the assertion re-derives fresh at the next intake.

A row whose current route cannot be established (no bundle **and** the deterministic pass returns no candidate) is **kept** — never drop on ignorance. Every decision, drop or keep, is printed with its reason.

## History is preserved

Dropped rows are appended to **`state/quarantine_resolved.csv`** (new, git-tracked, append-only — same spirit as `state/verdicts_applied.jsonl`) with the full original row plus `resolution`, `resolution_reason`, `current_candidate`, `resolved_by`, `resolved_date`. Nothing is silently discarded: a quarantine row records a real disagreement between two verifiers.

`README.md`'s state-file/issue table gains the new file, the assertion-verification block documents the step, and the "Running" section lists the command.

## What it dropped

Run in this PR, `quarantine.csv`: **2 → 0 open rows**.

```
DROP  ethiopia|fao1952|1937-1937  [route_changed]
      no longer routes to AOI-1936-1941; current route is ETH-1936-1941
DROP  ethiopia|iia|1938-1938      [route_changed]
      no longer routes to AOI-1936-1941; current route is ETH-1936-1941
```

Both were quarantined when no polity held Ethiopia proper during the 1936-41 Italian occupation — the two verifiers agreed on the substance (`fao1952` and `iia` both keep Ethiopia/Eritrea/Somaliland as separate labels, so "Ethiopia" is Ethiopia proper, not the AOI aggregate) but disagreed on the remedy. `ETH-1936-1941` has since been created and source-scoped aliases route both labels there, so the disagreement is moot. Verified independently by the reconciler's matchlib re-derivation (both keys now resolve to `ETH-1936-1941`). Their ledger status stays `issue`, so they re-verify and bank cleanly under the new family at the next intake.

## Verification

```
python3 scripts/validate_citations.py        PASS (0 missing sources/anchors)
python3 scripts/build_database.py --check    PASS (740 rows from 740 pages)
python3 scripts/update_wiki_index.py --check PASS
python3 scripts/validate_polygons.py         PASS
python3 scripts/audit_family_shadowing.py    PASS (0 shadowing pairs)
```

**Idempotency** — three checks:

1. Second consecutive run: `quarantine.csv is empty — nothing to reconcile`, exits without touching either file.
2. Re-quarantine simulation: copied a resolved row back into `quarantine.csv` and re-ran — it was dropped again, and the audit trail reported `archived 0 row(s) (1 already recorded)`, still 2 rows in `quarantine_resolved.csv`. Dedup ignores `resolved_date`, so a later-day re-run cannot duplicate it either.
3. Synthetic rows exercised the other two paths: a key with ledger status `fixed` → `DROP [banked]`; an unroutable key → `KEEP  route not resolvable`.

Closes #20


lbm364dl:
Reviewed independently — this holds up.

I ran the script from this branch against my own tree, where `state/assertions.json` **is** present, so it took the authoritative path rather than the `matchlib` re-derivation the worktree exercised. **Both paths agree**: each Ethiopia row is dropped as `route_changed`, from `AOI-1936-1941` to `ETH-1936-1941`. Two independent derivations of the same conclusion is better evidence than either alone.

`--dry-run` confirmed to write nothing.

The design call — a standalone script rather than a flag on `apply_verdicts.py` — is right for a reason worth stating: `apply_verdicts.py` loads a verdicts file and `assertions.json` at module scope, and **neither exists in a fresh checkout** (both gitignored per-run artefacts). Housekeeping has to work without them, which a flag on that script could not.

Also right: keeping rows whose route cannot be established, and leaving the ledger status at `issue` so the assertions re-verify and bank fresh rather than being assumed correct.

One limitation worth recording, not blocking: the `matchlib` fallback passes `iso=None`, since the assertion key carries only `label|source|span`. For labels that resolve by ISO in the real data, the fallback may return no candidate (→ row kept, safe) or resolve by name to somewhere else (→ a false `route_changed`). The consequence of a false drop is mild — the row is archived with its reason and the assertion re-verifies — but the authoritative `assertions.json` path should be preferred whenever available, which it is.

**This PR exposed a hole in my CI config**, now fixed on `main` (`50e0e11`): the `pull_request` trigger had a paths filter, so this PR got **"no checks reported"** and could have merged unvalidated. The filter is gone for PRs — an absent check is silent, which is worse than a failing one. Rebase on `main` and the suite will run here.

Issue #37, opened 2026-07-28
Inverse of check C: 6 polities declare polygon_status=unassigned while carrying geometry


lbm364dl:
`scripts/validate_polygons.py` check C fails when a page **declares** a polygon it does not have. The **inverse is unchecked**: a page can declare `polygon_status: unassigned` while the build attaches geometry anyway. Six polities are in that state, and the diagnosis splits cleanly in two.

**1. Five are self-inflicted, from the 2026-07-27 structural changes.** When rows were superseded or retired they had `polygon_status` set to `unassigned`, but their `polygon_source`/`polygon_feature_id` were left intact — so the build still resolves a feature for them:

| polity | wiki_status | attached geometry |
|---|---|---|
| `BRA-1800-2025` | superseded | 8,471,741 km² |
| `IDN-1800-1889` | superseded | 1,876,795 km² |
| `IDN-1889-1945` | superseded | 1,876,795 km² |
| `IND-1800-1893` | superseded | 4,652,138 km² |
| `RSFSR-1917-1991` | retired | 21,824,142 km² |

That was a conceptual error on my part: **`wiki_status: superseded|retired` already says "this row must not receive data"** (`matchlib.DEAD_STATUS` enforces it). Overloading `polygon_status` to repeat the message made the page assert something false about its own geometry. The honest encoding leaves `polygon_status` describing what is *actually attached* and lets `wiki_status` carry deadness. Keeping the geometry on a dead row is arguably useful for provenance — the row records what territory was previously used — so the fix is to restore the real status, not to strip the binding.

**2. `ADE-1839-1963` is a genuine live case.** It is `status: draft`, `polygon_status: unassigned`, yet carries a CShapes gwcode 680 geometry measuring **287,006 km²** against a claimed 290,000. So a live page understates what it has: anything filtering on `polygon_status` would wrongly treat it as polygon-less, and the 1% area agreement suggests the binding is fine and the status is simply wrong.

**Work:**
- Add the inverse check to `validate_polygons.py`: `unassigned` must mean *no geometry attached*. Fail on violation, since it is a direct contradiction like check C.
- Restore the five superseded/retired rows' `polygon_status` to what their binding actually produces.
[...]

Issue #6, opened 2026-07-27
Dangling predecessor/successor references — add a referential-integrity guard


lbm364dl:
`can-1948-2025`'s frontmatter lists `predecessor: [CAN-1866-1948]` — **a polity code that does not exist**. The chain's actual predecessor is `CAN-1886-1948`.

Nothing currently validates that `predecessor` / `successor` codes resolve to real rows, so there are probably more. Found by accident while investigating a routing change.

**Proposed guard** — a new check, either its own script or added to `scripts/validate_polygons.py`:

1. Every code in `predecessor` / `successor` exists in `polities_database.csv`.
2. Links are **symmetric**: if A lists B as successor, B lists A as predecessor.
3. Referenced codes are not `retired` / `superseded` (pointing a live chain at a dead row is how the retired-polity bug propagated).
4. Optionally: successor `start_year` equals predecessor `end_year`, flagging chain gaps and overlaps.

Point 4 would likely surface real dating problems — today's work already found a 225-year Brazil row hiding a territorial change, and a Canada row whose start year (1948, CShapes' referendum date) disagrees with the actual accession (31 March 1949).

Wiring it into CI alongside the other validators is the point; see the CI issue.

lbm364dl:
Related finding from the 2026-07-27 docs pass — **same class, different field.** Six pages used CSV *column* names where `build_database.py` expects *frontmatter* keys:

| page | wrong key | expected |
|---|---|---|
| `guf-1816-1946`, `guf-1946-2025`, `reu-1816-1946`, `reu-1946-2025` | `iso3_code:` | `iso3:` |
| the four above + `ptind-1816-1961` | `cow_code:` | `cow:` |
| the five above | `wiki_status:` | `status:` |
| `nrh-1911-1953` | *(no status field at all)* | `status:` |

The builder silently dropped them, so five polities had **no status in the database** and two had **no ISO code** — a value that looks present in the page but resolves to nothing, exactly like the dangling `CAN-1866-1948` reference this issue is about. Fixed in the docs-pass commit (`wiki_status` NaN count 6 → 0).

**Worth folding into this issue's guard:** as well as checking that referenced *codes* exist, validate that frontmatter *keys* are the expected set. Any key in a polity page's frontmatter that `build_database.py`'s `CSV_COLUMN_TO_FM_KEY` map does not recognise should fail — that catches both this typo class and future ones, and it is a two-line check given the map is already declared.

[...]

PR #38, opened 2026-07-28
Consolidate polygon_status to five values and close check C's blind spot

### LINKED ISSUE #31: Consolidate polygon_status: 9 values, 4 of them invisible to the validator
`polygon_status` has **9 values plus one NaN** in use, several of which are near-synonyms that grew organically. The vocabulary is now load-bearing — `scripts/validate_polygons.py` keys off specific strings — so the sprawl is a correctness risk, not just untidiness.

| value | count | intended meaning |
|---|---|---|
| `assigned` | 639 | the polygon **is** the territory (the only value check A fails on) |
| `proxy` | 36 | knowingly inexact stand-in |
| `unassigned` | 26 | no polygon, reason documented |
| `estimate` | 13 | approximate, no exact feature exists |
| `derived` | **9** | ← near-synonym of what? |
| `missing` | **9** | ← near-synonym of `unassigned` |
| `polygon_vintage_drift` | 4 | vintage outside the row's span |
| `approximate` | **2** | ← near-synonym of `estimate` |
| `excluded` | **1** | ← distinct meaning, or another `unassigned`? |
| *(NaN)* | 1 | no value at all |

**The concrete risk.** `validate_polygons.py` check C fails when a status *asserting a polygon* is set but none attached. Its set is `{assigned, proxy, estimate, polygon_vintage_drift}` — so **`derived`, `missing`, `approximate` and `excluded` are invisible to it.** A page marked `derived` with no geometry passes silently. The 9 `missing` pages are presumably fine (the name suggests no polygon is expected) but nothing enforces that reading, and `derived` almost certainly *should* have geometry.

**The pages, so this is checkable rather than abstract:**

- `derived` (9) — all aggregates/regions: `ANT-1961-2010`, `BLX-1850-1999`, `RAFR/RASI/REUR/RNAM/ROCE-1850-2021`, `RLAM-1850-2013`, `ROW-1850-2023`. Likely means "built by a builder", i.e. what `assigned` means for a `constructed` source.
- `missing` (9) — `AOI-1936-1941`, `FCC-1862-1887`, `FTO-1920-1960`, `NUP-1800-1897`, `RWB-1919-1922`, `SAA-1947-1957`, `TAN-1891-1920`, `TAN-1920-1922`, `TAS-1825-1900`. Probably `unassigned`.
- `approximate` (2) — `HND-1800-2025`, `MMR-1885-2025`. Probably `estimate`.
- `excluded` (1) — `GCO-1884-2025`. Needs a look: deliberately out of scope?
- NaN (1) — needs a value.

**Proposed target vocabulary** (5 values, each with defined validator behaviour, already documented in `wiki/README.md`):

```
assigned | proxy | estimate | polygon_vintage_drift | unassigned
```

**Migration:** `derived` → `assigned` (they are built polygons and should be area-checked), `missing` → `unassigned`, `approximate` → `estimate`, `excluded` → decide per page, NaN → fill. Then narrow check C's set to exactly the four asserting values and delete the legacy names from `wiki/README.md`'s do-not-use note.

21 pages plus the validator. Small, but it must be done as one change so the checks and the data stay in step — which is why it was left out of the 2026-07-27 docs pass rather than done as a drive-by.

lbm364dl:
`polygon_status` had **nine values plus one page with none**. Four of them —
`derived`, `missing`, `approximate`, `excluded` — were near-synonyms of the five
documented in `wiki/README.md`, and **none was in `validate_polygons.py`'s
`CLAIMS_POLYGON` set**, so 21 pages (plus `DJI-1886-2025`, which carried no value
at all) were exempt from check C without saying so: any of them could declare a
polygon and carry none, and nothing would fail. That blind spot is the actual bug
here; the tidying is secondary.

## Migration applied

Verified page by page against `data/final/polities_database.gpkg` rather than by
string substitution.

| was | n | now | verification |
|---|---|---|---|
| `derived` | 9 | **`assigned`** | all nine are `reporting-areas` aggregates and all nine **do** carry geometry — measured in an equal-area projection, 795 km² (`ANT-1961-2010`) to 2,569,652 km² (`ROW-1850-2023`); `BLX-1850-1999` at 33,344 km² against the 33,114 km² sum of modern Belgium + Luxembourg. `derived` described *how* the polygon was built, not what it claims. `polygon_area_km2` is null on all nine, so check A stays inactive; check C now covers them. |
| `missing` | 9 | **`unassigned`** | confirmed none carries geometry, so each is a documented gap rather than a claim: `AOI-1936-1941`, `FCC-1862-1887`, `FTO-1920-1960`, `NUP-1800-1897`, `RWB-1919-1922`, `SAA-1947-1957`, `TAN-1891-1920`, `TAN-1920-1922`, `TAS-1825-1900` |
| `approximate` | 2 | **`polygon_vintage_drift`** | `HND-1800-2025`, `MMR-1885-2025` — deviates from the issue, see below |
| `excluded` | 1 | **`unassigned`** | `GCO-1884-2025` carries no geometry; "excluded" named a scope idea the vocabulary does not track, and the row is itself a deletion candidate (`end_year=2025` for German Pacific colonies) |
| *(none)* | 1 | **`unassigned`** | `DJI-1886-2025`, retired: CShapes feature 522 was deliberately reassigned to `FRS-1884-1977` on 2026-07-01, so a polygon here would double-count the territory. Also given `polygon_source: none`. |

Each changed page carries a short `## Polygon status consolidated (added
2026-07-28)` section stating what changed, why, and the measured area where one
exists. One `lint` entry added to `wiki/log.md`
(`lint-consolidate-polygon-status`).

## Where I deviated from the issue

**`approximate` → `polygon_vintage_drift`, not `estimate`** on both pages. The
issue proposed `estimate`; the task asked me to check whether each page documents
an approximation or a *vintage* problem specifically. Both document vintage:

- `HND-1800-2025` — the geometry is fine (**112,280 km²** measured against
  Biger 1995's 112,044 km², 0.2%). The page's stated reason is entirely temporal:
  one **1886** CShapes polygon spread over 1800-2025, predating the 1906 King of
  Spain arbitration and the 1992 ICJ ruling, and back-projecting across 1800-1886.
- `MMR-1885-2025` — geometry also fine (**675,971 km²** against the ~680,000 km²
  the page cites). The page already calls its own problem an "82-year vintage
  drift": a 1967 Cliopatria snapshot applied from 1885.

That is the same use as the existing `polygon_vintage_drift` pages
`BRA-1800-1903` (1890 polygon, row from 1800) and `IDN-1800-1945` (1900 polygon,
row from 1800) — note that in existing practice the vintage sits *inside* the
span, so the value marks a vintage unrepresentative of much of the span rather
than one strictly outside it. Validator behaviour is identical for `estimate`
and `polygon_vintage_drift` (divergence reported, not failed), so the more
specific label was chosen. Easy to flip to `estimate` if you prefer the
narrower reading of the README wording.

## Validator

`CLAIMS_POLYGON` is now derived from a single `VOCABULARY` constant
(`VOCABULARY - {"unassigned"}`) instead of a hand-kept list — the hand-kept list
is what drifted. `EXACT_CLAIM` stays `{"assigned"}`, now documented as the
complement of the inexact values.

Added **test V**: fails on any `polygon_status` outside the five, including an
empty one. The migration closed today's blind spots; test V is what stops the
tenth synonym from re-opening them. Also updated `wiki/README.md` (the "legacy
values" paragraph now records the migration as complete and points at test V)
and `wiki/polities/_template.md`, whose comment still listed
`assigned | proxy | missing | excluded`.

## Verification

All green on this branch:

```
python3 scripts/build_database.py            # 740 rows, 689 geometries
python3 scripts/build_database.py --check    # PASS
python3 scripts/validate_citations.py        # PASS
python3 scripts/update_wiki_index.py --check # PASS
python3 scripts/validate_polygons.py         # PASS: 0 off-vocabulary, 0 area, 0 NEW claimed-but-absent
python3 scripts/audit_family_shadowing.py    # PASS
```

Check C still reports its 20 baselined claimed-but-absent rows and 0 new ones;
no baseline entry became stale.

**`data/final/` was rebuilt** and is committed, since frontmatter changed: 22
rows differ in the CSV. The GeoPackage is byte-different but geometry-identical
(same 689 features, identical total area) — SQLite output is not reproducible
byte-for-byte. If another PR touching `data/final/` merges first, this branch
will need a rebuild rather than a textual merge of the `.gpkg`.

`site/` was not regenerated: `site/wiki/` is already behind `wiki/` on `main`
and is refreshed by `scripts/rebuild.sh` as a separate pass.

Closes #31


lbm364dl:
Reviewed — verified the migration against the GeoPackage independently rather than reading the diff:

- `derived` → `assigned`: **9 rows, all 9 carry geometry** ✓ so `assigned` is correct and check C now covers them
- `missing` → `unassigned`: **9 rows, 0 carry geometry** ✓
- `approximate` → `polygon_vintage_drift`: both HND and MMR do carry geometry ✓

Deriving `CLAIMS_POLYGON` from a single `VOCABULARY` constant, and adding **test V** to fail on any status outside the five, goes further than the issue asked and is the right instinct: the hand-kept list is precisely what drifted into 9 values in the first place, so without test V the next synonym silently re-opens the same blind spot.

Two things worth recording rather than changing:

**1. Your `polygon_vintage_drift` observation is a documentation bug, not a labelling one.** You are right that in practice the value marks an *unrepresentative* vintage rather than one strictly outside the span — `BRA-1800-1903` and `IDN-1800-1945` both have vintages inside their spans. The README wording I wrote ("vintage year sits outside the row's span") is too narrow and describes only the common case. Please widen that table entry in this PR if you are touching it anyway; otherwise I will follow up, since the definition should match how five pages actually use it.

**2. This PR interacts with #37, filed after you started.** Six polities declare `unassigned` while *carrying* geometry — the inverse of check C, which nothing catches. Five are my own error from yesterday's structural changes (superseding a row set `polygon_status: unassigned` even though `wiki_status: superseded` already means "do not use"). Your migration is compatible: the 9 rows you move to `unassigned` have no geometry, so they will not trip the inverse check when it is added. #37 notes it should land after this so the vocabulary is settled first.

On the merge mechanics you flagged: agreed that a GeoPackage conflict needs a rebuild rather than a textual merge, and note `main` now has a guard (`build_database.py` refuses to write a GeoPackage with fewer geometries than the existing one) — the #6 agent watched an unguarded worktree rebuild shrink it 12.6 MB → 208 KB. Rebase to pick that up, along with the CI fix that makes checks run on every PR.

lbm364dl:
## Your `approximate` → `polygon_vintage_drift` call is right; I have aligned #39 to it

#39 independently migrated the same four legacy statuses (I did not check the open PRs first — my mistake, noted on that PR). On three of the four we agree. On `approximate` we differed, and **you are correct.**

I chose `proxy`, reasoning from the README's definition of the alternative — "the polygon's vintage year sits outside the row's span" — and observing that both rows' vintages are *inside* their spans (CShapes 1886 within 1800-2025; Cliopatria 1967 within 1885-2025), so `polygon_vintage_drift` appeared not to apply.

Your note that "in existing practice the vintage sits *inside* the span" is verifiable and I confirmed it. **All four** rows already carrying the value have the vintage within their own dates:

```
BRA-1800-1903   span 1800-1903   vintage 1890
IDN-1800-1945   span 1800-1945   vintage 1900
IND-1800-1886   span 1800-1886   vintage 1880
POL-1919-1920   span 1919-1920   vintage 1919
```

So the value marks a vintage *unrepresentative of much of the span*, which is exactly these two rows' problem. You checked established usage; I checked only the wording.

**#39 now uses `polygon_vintage_drift` on both rows**, so the semantic disagreement between the PRs is gone. I also corrected the README wording that caused it — it now says the vintage is usually inside the span, cites two existing rows, and records that the old phrasing mislabelled two rows. That correction is in neither PR's original scope and is worth keeping whichever merges first.

**Trial merge result**, so you know what resolving costs: 7 files conflict, and with the semantic one settled the rest are mechanical — `data/final/polities_database.{csv,gpkg}` are regenerable, `wiki/log.md` is two appended entries, and `_template.md` plus `scripts/validate_polygons.py` are independent edits (mine adds check A0 for "declares no polygon yet has one" and a stale-baseline direction; yours derives `CLAIMS_POLYGON = VOCABULARY - {"unassigned"}`, which is the better structure and should win).

I still recommend **merging this PR first** and rebasing #39 on top, dropping its now-redundant migration commit.

lbm364dl:
## Closing: split in two, and both halves are now on main

This PR's data half — consolidating `polygon_status` from nine values to the documented five — reached main independently. Verified: main carries exactly `assigned` (648), `unassigned` (38), `proxy` (37), `estimate` (10), `polygon_vintage_drift` (6), plus one page with none.

That one page is `DJI-1886-2025`, which is `retired`, so it is exempt from the checks by design and harmless.

**The guard did not land**, and that was this PR's actual point — *"that blind spot is the actual bug here; the tidying is secondary"*. Ported as check **V** in #64, now merged:

```
V. VOCABULARY - 0 live polit(ies) carry a polygon_status outside the documented set
```

Ported rather than merged because this branch has 6 conflicts and predates main's A0, B and D checks plus the reachability reporting — taking the file wholesale would have deleted 107 lines of newer work.

Two things #64 adds beyond what was here: check V **gates** the exit code (a guard that prints and exits 0 cannot stop the field regrowing), and it is **mutation tested** in both directions — `ADE-1839-1963` set to `derived` gives exit 1, restoring gives exit 0.

Nothing from this PR is lost. Closing.

PR #36, opened 2026-07-28
Codify the structural-change checklist: catch silently orphaned rows

lbm364dl:
Structural changes to the polity set — create, split, merge, retire, re-date — re-route data rows invisibly. The matcher resolves a label by alias → iso3 family + year containment → name, so an edit can move rows that were never mentioned on the page being edited. All four failures in the issue were invisible in the wiki diff.

## What this adds

`scripts/structural_change_check.py`, a snapshot/compare tool:

```bash
python3 scripts/structural_change_check.py --snapshot   # BEFORE the edit
#   ... edit the wiki page(s), run scripts/build_database.py ...
python3 scripts/structural_change_check.py --compare    # AFTER
```

`--snapshot` re-runs `01_match_and_findings.py` and records per-polity matched-row counts, the total, the set of resolving `(entity, iso)` labels, and every `polygon_area_km2`. `--compare` re-runs the matcher and reports:

1. **TOTAL matched rows** before vs after — a drop is a FAIL (rows orphaned).
2. **PER-POLITY movement**, biggest first — the Newfoundland case, where the total held and 54 rows sat on the wrong side of a boundary. Reported as REVIEW: only a human can say the movement was the intended one.
3. **Polities that received rows and now receive zero** — a row superseded without its data re-routed.
4. **Labels that became UNMATCHED** — the Indonesia rename that orphaned 23 name-resolved rows.

then the mechanical follow-ups: `audit_family_shadowing.py` (the Hyderabad/type-rank case), `validate_polygons.py`, and a placeholder-area scan (exact multiples of 100,000 — the `ETH-1936-1941: 1000000` case). Only a round area *this change* introduced or altered blocks; pre-existing round figures are reported as separate cleanup, so the tool measures the change rather than the backlog.

The eleven-step checklist itself is in `pipelines/polity-autoimprove/README.md` — it lives with the pipeline because its central check needs that pipeline's matcher and the layer-B dataset — mapping each step to script-or-human, and keeping the human-only steps explicit: writing the page (frontmatter *and* body), marking superseded rows, the FAOSTAT-era rematch, and the `wiki/log.md` `decision` entry **naming who signed off**. `wiki/README.md` points at it from the page-editing side.

## Deliberately not in CI

[...]

PR #35, opened 2026-07-28
ITS-1908-1960: composed-difference polygon (CShapes 520 minus 521), 464,743 km²

lbm364dl:
## Problem

`ITS-1908-1960` (Italian Somaliland) was bound to CShapes 2.0 gwcode **520 @1960** as a `proxy`, measuring **636,242 km²**. That feature starts on **1960-07-01**, so it is the whole of *independent Somalia* — Italian Somaliland **plus** British Somaliland after their union on that date. It overstated this row by roughly **137,000 km² (~21%)**, with **103 layer-B rows** matched here.

## Fix — composed difference

Added `build_its_1908_1960()` to `scripts/sources/constructed/build.py`, following the `build_chn_1932_1945()` (China minus Manchukuo) precedent:

```
CShapes 2.0 gwcode 520 @1960 (Somalia, 636,242 km²)
  MINUS gwcode 521 @1960 (British Somaliland, 171,499 km²)
  =                        464,743 km²
```

**Measured area: 464,743 km²** (from the built geometry, `ESRI:54034` equal-area). The simplified gpkg copy measures 464,286 km² (0.1% from simplify=0.01), well inside the validator's 25% tolerance.

`wiki/polities/its-1908-1960.md` now carries `polygon_source: constructed`, `polygon_feature_id: ITS-1908-1960`, `polygon_area_km2: 464743`.

## Status chosen: `assigned`

Not `estimate`, because the subtraction turns out to be **lossless rather than approximate**, and that is checkable:

1. gwcode 521 is **entirely contained** in gwcode 520 — their intersection equals 521 to the square metre (171,499.22457903888 vs 171,499.22457903856 km²). So the difference removes exactly the British share and nothing else.
2. The result is **bit-identical** to CShapes 2.0's own `Italian Somaliland` feature — cowcode 5200, time-steps from 1924-07-01, **464,743.0 km²** — with a **symmetric difference of 0.0 km²**. That feature exists only in the dependencies=TRUE distribution (`cshapes2_full.gpkg`), which is not one of the registered polygon sources in `scripts/sources.yaml` (the registered `cshapes-2.0` entry is the sovereign-state shapefile keyed on Gleditsch-Ward codes, and it has no pre-1960 Somali feature at all).

So this is not an inference about where the colonial boundary ran — it is CShapes' colonial boundary, reconstructed losslessly from the sovereign-only distribution. `assigned` is the honest label.

[...]

PR #34, opened 2026-07-28
validate_references: gate frontmatter keys, chain refs and codes named in page bodies

lbm364dl:
One new validator, `scripts/validate_references.py`, covering #6, #25 and the frontmatter-key check from #6's comments. They are the same sweep over frontmatter plus body, so they are one script. Wired into `.github/workflows/validate.yml` as a sixth step.

## What the check does

**1. Frontmatter keys** — every key must be one `build_database.py` reads. The legal set is re-derived from that file's `CSV_COLUMN_TO_FM_KEY` map (read out of the source, not imported, so the check needs no GDAL), plus an explicit `DESCRIPTIVE_KEYS` set of the 20 keys pages legitimately use to document a decision (`polygon_method`, `polygon_vintage_note`, `superseded_by`, `retired_reason`, `sources`, …). A CSV column name used as a key is reported as that specific typo class rather than as "unknown".

**2. Dangling chain refs** — every `predecessor`/`successor` code must exist in `polities_database.csv`.

**3. Symmetry and dead targets** — reported as WARNINGS that do not fail, and the docstring and output both say so. On current data that is 323 one-sided links and 18 live chains pointing into a `retired`/`superseded` row. Almost all are legacy pre-split chains; gating on them would make CI permanently red and the fix is a data cleanup, not a code change.

**4. Codes in page bodies** — this needed care. Bodies name ~240 nonexistent codes *correctly*: deleted rows (`GEA-1884-2025 has been deleted`), rejected or deferred splits (`a split into DEU-1938-1939, DEU-1939-1942 … is deferred`), hypothetical future rows (`a future ALK-1959-2025 row`). So bare mentions are warnings. What **fails** is a code the page asserts exists:
- a code on a `Predecessor:` / `Successor:` line, and
- a markdown link to a `wiki/polities/*.md` page that is not there.

Fenced code blocks are skipped, and a `retired`/`superseded` page's banner codes (its own `redirect` / `superseded_by`) are exempt, since those deliberately name the rows that replaced it.

## What it found

| check | found |
|---|---|
| 1. frontmatter keys | 7 on 6 pages — `polity_type:` on `guf-1816-1946`, `guf-1946-2025`, `reu-1816-1946`, `reu-1946-2025`, `nrh-1911-1953`, `ptind-1816-1961`; `last_updated:` on `ptind-1816-1961` |
| 2. dangling chain refs | 18 |
| 3. symmetry / dead targets | 323 + 18 (warnings) |
[...]

PR #39, opened 2026-07-28
Publish a consumer manifest so downstream copies cannot drift silently

lbm364dl:
## Start here

**What this does.** Publishes a contract the WHEP R package consumes instead of embedding a drifting copy: `polities_database.{csv,gpkg}`, `faostat_area_polity_map.csv`, `label_alias_map.csv` and `polities_manifest.json`, plus the gates that keep them honest. Paired with **eduaguilera/whep#382**, which consumes it.

**What is verified.** 25 validators, and a 26th that proves the others can fail — 11 of them are shown to detect an injected defect and name it. All registered in CI and in the README runbook, which the self-test asserts structurally so a gate cannot be added without both.

**CI cannot run here.** GitHub Actions billing is blocked on this repository, so every check above is verified locally only. That needs an account-level fix and is the one thing on this PR I cannot do.

**A reviewer's shortest path:**

| to check | look at |
|---|---|
| what the contract promises and who reads each field | "Consuming this database" in `README.md` |
| what a consumer must NOT infer from it | the `observed_rows` and `iso3_code` subsections there — both exist because a consumer inferred wrongly and it cost a 13.7× error |
| that the gates can fail | `python3 scripts/selftest_gates.py` |
| the newest gate | `scripts/validate_local_iso_codes.py` — 56 of 276 `iso3_code` values are local identifiers no ISO-keyed join can match |

**Open decisions**, filed rather than guessed: [issue 55](../../issues/55) on how dissolved states' ISO3 should be treated (three approaches coexist today), and the consumer-side set in eduaguilera/whep#425, #414 and #419.

---

Publishes a contract so downstream copies of this database cannot drift silently. Paired with **eduaguilera/whep#382**, which consumes it.

## Why

The WHEP R package embeds a copy of this database (`data/polities.rda`, built by its `data-raw/table_mappings.R`). That copy had drifted:

| | |
|---|---|
| rows in the package | **603** |
| rows here | **740** |
| missing | **144** |
| gone from here but still in the package | **7** |
| **retired or superseded here** | **24** |

The last row was producing wrong output, not just stale output. 24 FAOSTAT area codes resolved to withdrawn polities — **area 21 Brazil** to the collapsed `BRA-1800-2025` rather than the three rows that replaced it at the 1903 Acre acquisition, **area 100 India** and **area 101 Indonesia** likewise. So the territorial work done in this repo was not reaching the analysis, and the splits it created were invisible downstream.

Nothing detected this because checking meant diffing whole tables across two repositories, which nobody does by hand.

## What this adds

`data/final/polities_manifest.json`, generated by `scripts/write_manifest.py`:

```json
{
  "identity_sha256": "9b5d0456d3b41175…",
  "counts": { "total": 740, "live": 714, "dead": 26 },
  "dead_status": ["retired", "superseded"],
  "dead_polity_codes": ["AGO-1816-2025", "ARG-1800-2025", …],
  "live_polity_codes": [...]
}
```

Two deliberate design choices:

**The dead set is stated explicitly**, not left implicit in a status column. A consumer cannot repeat the dead-polity mistake *by omission* if the contract names the rows it must exclude. `dead_status` mirrors `matchlib.Matcher.DEAD_STATUS`, so both sides of the boundary agree on what "must never receive data" means.

**The hash covers identity, not geometry** — `polity_code`, `polity_name`, `start_year`, `end_year`, `polity_type`, `iso3_code`, `cow_code`, `wiki_status`. Re-measuring a polygon should not invalidate every downstream copy; a changed date or status must. Without that split the contract would cry wolf on every polygon fix, and get ignored.

## Verification

`scripts/write_manifest.py --check` gates it in CI alongside the existing five checks. Tested both directions — a clean tree passes; tampering with the row count fails and reports the before/after counts plus the hash prefixes.

All six checks pass on this branch.

## Follow-up on the consuming side

whep#382 already refreshes its copy (740 = 740) and excludes dead polities (24 → 0 bad routings). What it still needs is a test asserting its embedded table matches **this** manifest, so the drift fails loudly rather than being rediscovered by inspection. That is the next commit there.

---

# Update: what the branch grew into

The description above covers the first two commits. The branch is now 14. The later
work came almost entirely from one method — **comparing two independent
representations of the same fact** — and that is what found the defects below,
none of which were visible by reading the code or the data.

## Contracts published

| file | what it is | gated by |
|---|---|---|
| `data/final/polities_manifest.json` | row counts, identity hash over resolvable fields, live/dead code sets, polygon-gap set | `write_manifest.py --check` |
| `data/final/faostat_area_polity_map.csv` | 281 mappings: which polity a FAOSTAT area's data belongs to, per year range | `write_faostat_area_map.py --check` |
| `data/final/label_alias_map.csv` | 662 aliases: what a source's country **label** means, per source and year | `write_label_alias_map.py --check` |

The last two exist because a consumer that rebuilds these mappings itself becomes a
second authority for the same question — which is how the WHEP package came to
attribute **118 area-years** of FAOSTAT data to polities that did not yet exist,
including 50 years of Sudan routed to the post-2011 row that excludes South Sudan.

## Defects found, and how

**Five aliases were silently INERT.** A broken alias does not fail — `matchlib`
drops it and the label just never matches. Two had an unquoted comma in a field,
which shifted every column and dropped the target (`china, taiwan province of` had
never worked); one targeted the bare prefix `ETH`, which is a family and not a
polity; one targeted `GNGU-1884-1914`, which does not exist (issue 42). I found
three by inspection and then wrote `validate_aliases.py`, which found five. **3 of
5 by reading, 5 of 5 by gating.**

**The two matchers in this repo disagreed on 3 of 281 areas** (`crosscheck_matchers.py`,
which loads `matchlib` with the `faostat` rules removed to avoid circularity):
Serbia 2006-2008 exists **twice** (#43), `"Malaysia"` 1961 resolves to **British
North Borneo** because four same-`iso3` candidates are all typed `national` so the
tie-break cannot discriminate (#44), and `"Viet Nam"` routes to South Vietnam alone
by one alias and to the combined entity by the other.

**`RUS-1991-2014` carries the USSR polygon** — CShapes feature 365, the same as
every `F228-*` row — measuring 21.8M km² against Russia's 17.1M, and geometrically
containing **12 live polities** including Kazakhstan, Ukraine and the Baltics
(#45).

**`IRN-1800-1828` is drawn as the United States.** Its `polygon_feature_id` says
`"United States of America"` (#46).

**`FRA-1871-1919` recorded 62.43 km²** — CShapes' area in **square degrees**,
written into a km² field, asserting France was smaller than Manhattan. Corrected to
the measured 532,305.

## Why the existing gates could not see these

Each finding exposed a structural limit, and each is now fixed:

- **`validate_polygons` check A skipped rows whose CLAIMED area was under 200 km².** A
  claim wrong by being *far too small* exempted itself. The filter now considers the
  measured area too — that is what surfaced France.
- **Check A is opt-in via a hand-entered field.** ~590 of 740 rows carry no
  `polygon_area_km2` and were never compared. Russia is in that blind spot.
- **Magnitude cannot detect position.** The USA in 1815 was ~1,586,287 km² and Iran
  ~1,621,564 — ratio 0.98 — so no area check could ever find Iran. The eight
  wrong-country polygons caught previously were all caught because their sizes were
  absurd.
- **`audit_family_shadowing` requires a ≥3× area ratio**, calibrated on Alaska
  (~470×). Malaysia-vs-North-Borneo is 1.8× and sails through (#44).

So `validate_spatial_containment.py` adds two checks that need **no reference data
and no hand-entered value**: containment catches a polygon that is too *big*,
continuity catches one in the wrong *place*.

## Also here

- **`GCO-1884-2025` retired** — "German colonies Oceania" with `end_year: 2025`.
  Retired rather than deleted so the two pages referencing the code in prose keep a
  resolvable target.
- **19 aliases** resolving the labels three WHEP datasets carry (`BZE`, `ROM`, `ZAR`,
  `Cape Verde`, `Swaziland`…), year-split where the data window crosses a
  transition. `FSU` and `Belgium-Luxemburg` are deliberately left unresolved after
  1991 and 1999 — the source reports them past the end of the territory they name.
- **The prefix convention is documented** in `wiki/README.md`. I set out to unify
  `ANG`→`AGO` etc. and stopped: `SUD-1899-1934` is 2,579,525 km² — Sudan
  *including* South Sudan — so `SUD`→`SDN` would have merged two different
  territories. A prefix names a territory, not a regime, and 84 live prefixes
  differ from their `iso3_code` deliberately.
- **`build_database.py` keeps the GeoPackage's attributes in sync** when its geometry
  guard declines to rewrite, and `--check` now verifies them. The two published
  artifacts had contradicted each other: the manifest (from the CSV) called GCO dead
  while the GeoPackage that consumers actually read still called it live.

## Verification

All eleven gates pass. Every baseline in this branch is **bidirectional** — a new
defect fails, and a defect that has been *fixed* fails until removed from the
baseline — so the lists shrink as issues 40 and 42-46 are closed rather than
quietly licensing regressions.

---

# Update 2: the polygon work

The addendum above covers the branch through the cross-checking commits. Since then
the raw polygon sources turned out to be present in the main checkout, so a faithful
rebuild became possible — 689 geometries, the same count as committed, with **zero
geometries changing area by more than 1%**, established before anything was touched.

That unlocked two defect fixes and five new polygons.

## Two wrong-country polygons fixed

**`RUS-1991-2014` was drawing the USSR.** 21,824,142 km² against Russia's 17,098,242,
geometrically containing **12 live polities** — Kazakhstan, Ukraine, Uzbekistan, the
Baltics. The cause was not a wrong feature: CShapes gwcode 365 has **six** time-steps
containing 1991, and the build prefers a step whose start equals the queried year — all
six qualify — so **shapefile order decided it**. `polygon_feature_year: 1992` is
contained by only the durable `1991-2014` step. Now 16,882,058 km².

**`IRN-1800-1828` was drawing the eastern United States.** Its
`polygon_feature_id` said `"United States of America"` in plain words. Rebound to
Cliopatria `"Qajar Dynasty"` at 1800 → 1,743,638 km², centroid lon 53.9 lat 33.1.

Neither was visible to the area check, and for opposite reasons: Russia records **no**
`polygon_area_km2`, so nothing compared it to anything; and the USA in 1815 was
1,586,287 km² against Iran's 1,621,564 — **a ratio of 0.98**. Magnitude cannot detect
position, and an opt-in check cannot see a row that opted out.

## Polygon backlog 20 → 13

First the 20 rows were **diagnosed**, because they recorded each feature id but not what
was blocking it, so every entry looked like the same problem:

| tag | rows | what it needs |
|---|---|---|
| `[compose]` | 5 | a builder; all five need GADM extracts absent locally (Canada, India, Indonesia) |
| `[no-id]` | 4 | a source and id identified, or an honest downgrade |
| `[extract]` | 3 | nothing but a fuller GADM download — `ALK`, both `RYU` |

Then seven were closed:

- **Two already had polygons and simply never named them.** `NRH-1911-1953` had *only*
  `polygon_status` in its frontmatter — CShapes gwcode 551 has a step covering exactly
  1911-1953. `GRC-1830-1881` named Cliopatria with a `null` id — it carries "First
  Hellenic Republic" across the span.
- **Five were built**, each verified against numbers its own page already carried:
  `BTL` (26,630 km², matching the recorded figure exactly), `TTPI`, `AEF` (2,495,221 vs
  2,495,219 — the sum of its own constituent table), `GBM` (132,040, exactly its
  tabulated total), `SYL`.

## What I did not do, and why

**`GCT-1919-1956`'s polygon is verified and deliberately withheld** (issue 47).
Attaching it makes `audit_family_shadowing` fail: GCT and `BTL-1920-1957` both carry
iso3 `GHA`, both are typed `colonial`, and they overlap at 9×, so family ordering rather
than type would decide which gets GHA-labelled data. The polygon does not create that
risk — it makes it measurable. Settling it means deciding how a composite reporting unit
should be typed, and the database is inconsistent there today.

**The other five `[compose]` rows were left alone.** Unlike the ones built, their
compositions have no corroborating figures on their pages, and reconstructing a
territorial extent from an unverifiable specification is guessing.

**No external data was downloaded.** The three `[extract]` rows need a ~1 GB GADM fetch;
they are precisely diagnosed and tagged instead.

## Two new spatial checks

`validate_spatial_containment.py` needs no reference data and no hand-entered value:

- **containment** — a polygon that swallows a coexisting polity (found Russia)
- **continuity** — consecutive periods of one family that do not overlap (found Iran)

Both are bidirectional, and between them they have made **four** of my own changes
justify themselves: three newly built polygons flagged as new containers (all
legitimate federations, now enumerated with reasons), and every baseline removal forced
rather than remembered.

---

## Gates added since this PR opened

Fifteen now guard the database, each because that class of error was **found in the data**.
The README documents all of them; these are the ones added here after the original scope:

| gate | what it caught |
|---|---|
| `validate_iso_codes.py` | `FRS-1977-2025` is modern Djibouti and carried `iso3: FRS` — not an ISO code at all, so nothing holding a country code could reach it. Also Sudan as `SUD`, colonial Angola as `ANG` |
| `validate_iso_collisions.py` | *(guard)* 59 pairs already share a code over overlapping years **by design**, since `iso3` groups by modern territory. The gate is that the set must not grow |
| `validate_period_overlaps.py` | four same-family pairs cover the same years, so a year-aware matcher must guess. `PER-1825-1909` duplicates two rows that already tile its span exactly |
| `validate_reporting_areas.py` | six GADM territories claimed by **two** aggregates each — Palau and the Northern Marianas sit in both Asia Other and Oceania Other. Hidden because the RoW union deduplicates |
| `validate_succession_geography.py` | `NWR-1900-1905` (Northwestern Rhodesia) listing its successor as Northern **Nigeria**, 4,000 km away |

## The `iso3` correction, and what it unlocked

`iso3` names the **territory**; the polity-code prefix names the **chain**. Conflating them
meant a colonial-era row could only be reached through its era-specific key. Three fields
corrected — `ANG-1905-1975`→`AGO`, `FRS-1884-1977`→`DJI`, `SUD-1956-2011`→`SDN` — with no
prefix changed.

Result: `crosscheck_matchers.py`'s `BASELINE_UNRESOLVED` shrank from `{7, 72, 248}` to
`{248}`. **Its bidirectional arm is what reported that**, failing with *"area 7 is baselined
as unresolved but now agrees"*. Angola's and Djibouti's colonial FAOSTAT mappings resolve by
ISO now — the functional goal of the approved colonial/modern prefix unification, reached
without renaming any polity code, which would have cascaded through aliases, the FAOSTAT map,
the wiki and the consumer's embedded data.

Two earlier readings of that baseline were wrong and are recorded in the code so they are not
re-derived: it is not "the four China components" (those resolve fine), and it was not "a
prefix the label does not lead to" — true only of 248. For 7 and 72 the obstacle was
**temporal**, not nominal.

## Published contract, extended

- `observed_rows` in `label_alias_map.csv` — how many source rows were actually seen for a
  label, 0 when it is merely mappable. **Eleven areas carry substantial data** (Bermuda 67,310
  rows, Faroe Islands 45,036, Cook Islands 42,137, Palestine 32,534) while each has its own
  live polity this map already targeted, and their data was being attributed to
  `ROW-1850-2023`.

  **What the consumer may do with that is narrower than this bullet first claimed.** It uses
  `observed_rows` to decide which areas get their own *polity* — and that half is shipped. It
  must NOT use it to decide which get their own numeric aggregation key: doing so inflates
  global `feed` **13.7×**, with the whole increase landing on one area. So the eleven still
  aggregate into rest-of-world, deliberately, and the attribution problem stays open in
  eduaguilera/whep#419 rather than being traded for a larger error. The README section on
  `observed_rows` carries the measurement.
- **Ten short-form labels** aliased — `United Kingdom`, `South Korea`, `Tanzania`,
  `Ivory Coast`, `Czech Republic`, `DR Congo`, `Lao`, `Moldova`, `North Korea`, `Eswatini`.
  Each had a canonical long-form sibling that already resolved, so the polity existed and
  only the routing was missing. Eswatini mattered most: it has its own commodity-balance data
  and resolved to nothing under either name.
- **`SRM` repointed** from `SUR-1975-2025` to `SCG-1992-2006`. `SUR` appears in the same
  source file as its own code, so Suriname held two sets of nitrogen rates and SRM's real
  entity held none. Identified by elimination: every ex-Yugoslav republic is present while
  SRB, SCG and MNE are all absent.

## Open decisions (issues, not blockers)

#48 ROW geometry scope and six double-claimed aggregate components · #49 four same-family
period overlaps (`PER-1825-1909` looks mechanically redundant) · #50 live reporting buckets
ending before the data does, 72 area-years resolving to nothing · #51 `SNI-1906-1913` sharing
a polygon with `SNI-1899-1906` though Lagos Colony merged in that year

---

## Aliases: 673 → 869

The consumer has datasets keyed by country **names** and by anachronistic ISO codes, and this
repository owns label-to-polity identity, so the mappings belong here rather than as name-matching
logic there.

| source | added | why |
|---|---|---|
| `lassaletta-grassland-share` | 158 | country names + year; consumer coverage 14.5% → **98.1%** |
| `trade-sources` | 34 | 8 historical reporters, spans back to 1809 |
| `crops-manure-n` | 4 | modern ISO codes for a circa-2000 snapshot |

**Rules the generation follows**, since "generated mechanically" is the part to distrust: exact match
to a `polity_name`; single live family only; dead rows excluded via `dead_status` **read from the
published manifest** rather than a sixth hardcoded copy; one alias per polity period, non-overlapping
and following the exclusive `end_year`.

Two of those rules exist because the first generation broke them. It emitted 5 rows targeting
**retired or superseded** polities — rows that may never receive data. They surfaced as *overlapping
year ranges*, which is how a dead duplicate of a live period looks from outside: two same-named
polities both spanning 1961–2009. `validate_period_overlaps` reports only 4 overlaps precisely
because it filters dead rows, and my generator did not.

**The 88% → 98.1% step came from a realisation, not new curation.** I had written that residue up as
"predecessor-era cases needing curation decisions" — when the `faostat` source already held those
decisions for the identical labels: Germany before 1990 as `DEU-1949-1990`, Yemen as `F249-1918-1990`,
Bangladesh as `BGD-1947-1971`, and Morocco's two-family ambiguity already settled as `MAR`. 28 of the
last 29 aliases are copies of decisions someone had already made.

**Cases that resolve to no area at all.** China's FAOSTAT area 351 is deliberately unmapped against
double-counting its components, so an area-mediated lookup returns nothing for the reporter with the
longest chain in the trade file. Its aliases target the `CHN` family directly, which this table
permits because it maps labels to **polities**, not areas.

## observed_rows: stop reporting "not measured" as "measured zero"

The column coerced an empty registry value to `"0"`. Two different facts, one encoding:

| state | meaning | aliases |
|---|---|---|
| a number | measured, that many rows | 384 |
| `0` | measured, genuinely none | 29 |
| empty | **not measured** — that source's corpus is not in this repo | 393 |

Every `lassaletta-grassland-share`, `mueller-synthetic-n` and `crops-manure-n` alias is in the third
state, because those datasets live in the consumer. Measured there they are the opposite of inert:
6,082 Lassaletta country-years resolve, 184 `crops_manure_n` codes, 156 Mueller codes. Publishing `0`
read as 152 dead aliases, and an inert-alias check on this column would have flagged every one.

Consumer impact was checked, not assumed: its FABIO folding filter already read
`!is.na(observed_rows) & observed_rows > 0`, so `NA` is excluded exactly as `0` was — confirmed by
rebuilding the crosswalk and diffing every `polity_area_code`, 0 of 218 moved.

## A check that proves the gates can fail

Validators that are all green are indistinguishable from validators that **cannot** fail.
`scripts/selftest_gates.py` mutates the GeoPackage, runs one gate, and requires a non-zero exit **and**
that the output names the injected code.

Its more valuable result was stopping a plausible fix. The containment gate normalises overlap by the
*smaller* polygon, so a successor bound to a wrong feature inside the right territory scores ~1.0 and
passes. Symmetrising is the obvious repair — and measuring first showed 26 consecutive pairs would be
flagged, overwhelmingly **real history**: `USA-1848-1867 → USA-1867-1959` is the Alaska purchase,
`ISR-1948-1967 → ISR-1967-1979` the Six-Day War, `HUN-1918-1919 → HUN-1920-1938` the Treaty of
Trianon. The smaller-normalised metric is invariant to legitimate expansion, which is why it is
correct; the hole is covered by `validate_family_areas`, which measures that same shrink at **222×**
its family median.

---

## Twenty-five gates, and eleven of them prove they can fail

Three were added late in the branch, each because a defect surfaced in the CONSUMER rather than here — a
contract's faults show up where it is consumed:

| gate | what the consumer hit |
|---|---|
| `validate_code_year_agreement.py` | two aliases resolved 1962–64 to a polity whose columns say it ended in 1961, because they were written against the years in its **code**. `TAN-1922-1964` has `end_year` 1961; `NNG-1949-1963` has 1969. Both are historical judgement (independence vs union; transfer vs Act of Free Choice), so baselined |
| `validate_alias_chain_overlaps.py` | 25 label chains where consecutive aliases both cover a boundary year, so match order decides which polity a value lands in. Rate depends on how they were written: **18 of 31** hand-entered any-source chains against **0 of 7** generated with `year_end = polity_end_year - 1` |
| `selftest_gates.py` | grew from three cases to eight |

The alias-overlap gate **gates rather than fixes**, deliberately: a mechanical pass would move 91 rows, each
shifting a boundary year's data, and one of the 91 is the Cape Verde mid-year-independence choice made on purpose.
A bulk edit would erase that decision along with the 88 accidents.

## The self-test earns its keep

Eight cases now, each requiring the gate to exit non-zero **and name the injected defect**. That second
requirement is what makes the harness trustworthy rather than decorative:

- case 8 (`build_database.py --check`) exited 1 **twice for the wrong reason** — once a `FileNotFoundError` from a
  missing `sources.yaml`, once because I expected the message to name a field when it names the differing row.
  Exit status alone would have recorded a passing case in both states: a self-test that believes a crash is a
  detection.
- case 7 needed `--check` in its argument list, because `write_manifest.py` bare **regenerates** the manifest from
  the mutated CSV and exits 0 — absorbing the defect it was meant to detect. Verified against a fresh copy per
  mode, after a first harness that ran write mode and `--check` in the same directory and so repaired the defect
  before measuring it.
- case 5 exists because its own gate shipped with a blind spot: it skipped rows with empty year bounds and so
  missed `turkey`, where an unranged alias overlaps a ranged one at every year. That was found by **reconciling
  the gate's count against the consumer's** — 24 against 25 — not by mutation. A self-test proves a gate fires on
  the defect you thought of, which is necessary and not sufficient.

## observed_rows, and what a contract owes a consumer

`observed_rows` coerced an empty registry value to `0`, so 393 aliases read as "measured, no data" when they were
"not measured here". Every `lassaletta-grassland-share`, `mueller-synthetic-n` and `crops-manure-n` alias is in
that state, because those datasets live in the consumer — and measured there they are anything but inert: 6,082
Lassaletta country-years, 184 `crops_manure_n` codes, 156 Mueller codes. The consumer now runs the inertness check
those 152 aliases need, since this repository cannot.

## A source scope is not a spelling

Four aliases — `DR Congo`, `Ivory Coast`, `Lao`, `Czech Republic` — were added for the harmonization tables and
scoped to `source = faostat`. The labels they carry are ordinary English short forms, so the scope made them
unreachable for every other consumer, including one that resolves areas by name with no source at all.

Measured downstream, on WHEP's `crop_residues` (475,688 rows), those four accounted for **9,822** of the rows still
carrying no area code once the consumer learned to route unmatched names through this database. Rescoped to blanket,
alongside the seven `czech republic` rows that were already blanket, they resolve. What remains is 200 rows — all
Tanzania 1961-1964, blocked by the `TAN-1922-1964` code/column disagreement that `validate_code_year_agreement.py`
already baselines. **That baseline now has a measured downstream cost** rather than an assumed one.

## A bound nothing reads is not a bound

`validate_unranged_aliases.py` permits `italy | iia | (blank) | 1860 -> SAR-1800-1860` on the stated grounds that
its `year_end` is "bounded exactly where it matters", because the rule it enforces is about the upper bound alone.
That was true of the gate and **false of both matchers**:

```
matchlib.match_alias:  if ru["y0"] is not None and not (y0 <= year <= y1): continue
```

With no `year_start`, the year test is skipped entirely and the rule matches every year — so IIA data labelled
"italy" resolved to Sardinia in **the year 2000**, to a polity that ended in 1860. WHEP's `resolve_polity_label()`
had the same shape, which is where it surfaced: a consumer in another repository hit an answer this repository's own
gate had reasoned was impossible.

A missing bound is now unbounded on that side — in `match_alias`, in the blanket classification, and in the
ambiguity guard, which had been skipping half-open rules rather than comparing them. `italy`/iia now gives
`SAR-1800-1860` for 1850 and 1860, `ITA-1861-1866` for 1861, and nothing for 2000.

Exactly one of the 869 published aliases carries a single bound, so this moves one row. The point is the class: two
components of one repository read one field differently, and the gate permitting the row was the one that had it
right.

Also: `write_manifest.py --check` reported this change as `869 -> 869 aliases`. The count was identical and the
digest was not, so the one number it printed was the one number that had not moved. It now names whichever subfield
actually differs — the same rule the polity counts above it already followed, for the same reason.

## Guarding a property the consumer depends on

`resolve_polity_label()` in WHEP now resolves a label by alias first and then by the polity's
own name, mirroring `matchlib.Matcher`'s name family. Both sides can only answer when
**exactly one** polity of that name is live in the year asked about — otherwise they return
nothing, because picking by row order would invent an answer. So "one live polity per
(normalised name, year)" is a property the consumer depends on, and nothing here guarded it.

The failure mode is silent downstream: renaming a polity so its normalised name collides with
a live sibling breaks no join and raises no warning. It turns a label that used to resolve
into `NA`, and `NA` is what an unmapped label looks like too.

**Measured:** 732 live polities, 110 normalised names held by more than one, and **17** whose
holders overlap in years. Sharing a name is fine — Greece 1830-1881 and Greece 1947-2025 are
one country in two eras and a year separates them. Sharing a name *and* a year is what cannot
be resolved.

I wrote in the new gate's docstring that all 17 were the same shape, and running it disproved
that:

| count | shape | tracked as |
|---|---|---|
| 15 | a coarse period listed beside its own finer sub-periods, same prefix (`BLZ-1800-2025` spanning `BLZ-1800-1886`, `BLZ-1970-1981`, `BLZ-1981-2025`) | issue #49 |
| 1 | `morocco`: `MAR-1911-1958` against `MOR-1956-1958` — crosses prefix families | issue #52 |
| 1 | `serbia`: `SER-2006-2008` against `SRB-2006-2008` — a duplicate, not a periodisation | issue #43 |

Baselined bidirectionally, so the list can only shrink deliberately and the two cross-family
cases leave it when they are fixed where they are tracked.

**The self-test case earned its keep before the gate guarded anything.** It declared the wrong
file writable, so `stage()` left the CSV as a symlink and the mutation wrote *through it into
the real database*. `validate_cross_family_names` reported the identical injected defect
independently within seconds — which is the best evidence either gate works, and a reminder
that the harness's own comment ("those cases ask for a real copy via `writable`") was there to
be read rather than pattern-matched around.

Twenty-five gates now, eleven of them shown to fail on an injected defect and to name it. The newest baselines the `iso3_code` vocabulary, because 56 of its 276 values are local identifiers no ISO-keyed join can match, and that was documented in prose and guarded by nothing.


lbm364dl:
## Overlap with #38 — read before merging

I duplicated **#38**'s work. It has been open the whole time, does the same `polygon_status` migration this branch does, and identifies the same root cause (the four legacy values sitting outside `CLAIMS_POLYGON`, so 21 pages were exempt from check C without saying so). I did not check the open PRs before migrating 21 pages. The two now conflict.

**#38 is the better piece of work on the overlapping part:**

- It fixes `wiki/polities/_template.md`, which I had missed. That mattered: my branch left the template documenting `assigned | proxy | missing | excluded` with an example of `polygon_status: missing` — a value my *own* new vocabulary gate rejects. The gate and the template disagreed. (Fixed here now too, but #38 got there first.)
- It gives `DJI-1886-2025` a value; this branch leaves it empty.
- It derives `CLAIMS_POLYGON = VOCABULARY - {"unassigned"}` **in one file**, so the two cannot drift. Structurally better than what I did.

**What this branch has that #38 does not:**

- `scripts/validate_constants.py`, whose more valuable half checks that **`DEAD_STATUS` agrees across all five files that define it** (`write_manifest`, `validate_polygons`, `validate_aliases`, `validate_spatial_containment`, `matchlib.Matcher`). A drifted copy there would let the matcher route data to a row the published contract calls dead.
- A `polygon_status` vocabulary gate in `build_database.py --check`.
- Everything else in this branch — the published contracts, the two spatial checks, the polygon fixes.

**Suggested order:** merge **#38** first, then rebase this branch. The migration commit here becomes redundant and should be dropped in the rebase; `validate_constants.py` survives, with its `CLAIMS_POLYGON` half simplified (once #38 derives it, checking it against the vocabulary is trivially true) and its `DEAD_STATUS` half intact.

**#35** (`ITS-1908-1960` composed-difference polygon) also conflicts, on `scripts/sources/constructed/build.py` — this branch added six builders to that file. That one is a straightforward textual merge; both sets of builders are independent.

lbm364dl:
## Correction: I under-reported the chain-reference problem

Two commits ago I investigated `predecessor`/`successor` frontmatter, found 239 asymmetric links, established that **nothing** in `scripts/` or `pipelines/` reads those fields, and recorded the conclusion that they were inert metadata — "not a defect; churning 239 pages would be motion without value".

That was wrong about a subset. Re-checking with the categories separated:

- **19 links point at a polity code that does not exist** — `CAN-1800-1866 → CAN-1866-1948` (the real code is `CAN-1886-1948`), `BGD-1947-1971 → PAK-1940-1947`, `CHL-1899-1902 → CHL-1810-1899`, and 16 more.
- **20 links point at a `retired` or `superseded` row.** I never checked this category at all.
- 235 are genuinely asymmetric with both codes real — those *are* benign, as I said.

I folded "the referenced code does not exist" into the same bucket as "the link is one-directional" and reported a single number, which made a real defect invisible inside a benign one. The distinction that matters is not symmetry, it is whether the reference **resolves**.

**#34 already fixes this**, and reaches the same conclusion about symmetry independently (it reports asymmetry as a warning that never fails, because pre-split chains legitimately produce it). Its diff touches the same pages my dangling list names. I have made no changes here — duplicating it would repeat the #38 mistake.

lbm364dl:
## Merge guidance across all seven open PRs

I audited this branch against every other open PR here, after discovering I had duplicated #38. Summary for whoever merges.

### Overlaps with this branch

| PR | relationship | action |
|---|---|---|
| **#38** consolidate-polygon-status | **Duplicates this branch's `polygon_status` migration.** #38 is the better version: it also fixes `_template.md`, gives `DJI-1886-2025` a value, and derives `CLAIMS_POLYGON = VOCABULARY - {"unassigned"}` in one file so the two cannot drift. | **Merge #38 first**, then rebase this branch and drop its migration commit. `validate_constants.py` survives — its `DEAD_STATUS` half (five definitions, not in #38) is the part worth keeping. |
| **#34** validate_references | **Catches a defect I wrongly dismissed** — 19 dangling chain refs and 20 pointing at dead rows, which I folded into a single "239 asymmetric, benign" number. Independently corroborated in a comment there. | Merge. No changes needed here. |
| **#35** its-somaliland-polygon | Textual conflict only, on `scripts/sources/constructed/build.py` — this branch added six builders to that file, #35 adds one. Both independent. | Straightforward merge either order. |
| **#32** quarantine reconciliation | No conflict. Its Ethiopia reconciliation routes to `ETH-1936-1941`, which exists and is live in this branch (`ETH-1907-1952` is correctly `superseded`). Verified all 662 aliases here target live polities. | Independent. |
| **#33** protocol-version-reopen | Pipelines only. No overlap. | Independent. |
| **#36** structural-change-check | Both touch `wiki/README.md`, but different sections — #36 does not mention the "Prefix convention" section this branch adds (0 matches). | Independent; glance at the README merge. |

### Suggested order

1. **#38** — then rebase this branch, dropping the redundant migration.
2. **#34** — independent, and fixes the chain refs.
3. **#39** (this one) — after the #38 rebase.
4. **#32**, **#33**, **#35**, **#36** — independent of the above, any order.

### One thing to check at merge time

Both **#34** and this branch add steps to `.github/workflows/validate.yml`. This branch takes it to **14** steps, adding `validate_constants.py`, `validate_spatial_containment.py`, `crosscheck_matchers.py`, `write_label_alias_map.py --check` and `write_faostat_area_map.py --check`. Git reports both mergeable, so the step lists reconcile, but the resulting order is worth a look — this branch deliberately runs the FAOSTAT-map check *before* the manifest check, because the manifest fingerprints that map and the specific failure is more useful than the general one.

### And for the paired PR

**eduaguilera/whep#382** should merge *after* this one. It reads five files from `data/final/`; two are already on `main` here, and the other three — `polities_manifest.json`, `faostat_area_polity_map.csv`, `label_alias_map.csv` — arrive with this PR. Only `label_alias_map.csv` is needed to *build* rather than to test, so between the two merging it is the one artifact a `data/` regeneration can be missing; that failure now explains itself rather than reporting a bare path.

Issue #195, opened 2026-08-10
Half the declared polygon areas are read off their own geometry, so the check comparing them is tautological


lbm364dl:
> *Filed by Claude (Claude Code). Every figure measured against the current database. Where a case is my own doing I say so.*

`validate_polygons` check A compares each row's declared `polygon_area_km2` against the geometry attached to it. For at least half the rows that comparison **cannot fail**, because the declared figure was derived from that same geometry.

## Measured

197 live rows carry both a declared area and a polygon:

```
agreeing within  0.1%: 103  (52%)
agreeing within  0.5%: 132  (67%)
agreeing within  1.0%: 145  (74%)
agreeing within  2.0%: 155  (79%)
agreeing within  5.0%: 173  (88%)
```

Eight agree to **0.000%** — the same number to every digit:

```
FRA-1871-1919   532,305    cshapes-2.0
MAN-1932-1945   791,708    constructed
MAN-1945-1950   791,708    constructed
FID-1887-1954   753,049    constructed
HYD-1724-1948   211,298    cliopatria
SRB-2008-2025    77,251    cshapes-2.0
POL-1919-1920   177,754    cshapes-2.0
EGY-1820-1885 1,834,433    cliopatria
```

A polygon cannot disagree with a number copied from it. For those rows check A is a no-op that reports PASS.

## Why this is not merely inelegant: it has hidden three real errors

Each of these had a declared area agreeing with its geometry to under 1%, and in each case **both were wrong in the same direction**:

| row | declared | measured | agreement | what both included |
| --- | --- | --- | --- | --- |
| `IDN-OTH-1949-1951` | 1,757,495 | 1,747,408 | 0.6% | West Papua, Dutch until 1962 (#182) |
| `IND-1800-1886` | 4,209,917 | 4,209,869 | 0.001% | Ceylon, a separate Crown Colony (#191) |
| `CAN-1800-1866` | 1,209,852 | *(none built)* | — | the recipe with Québec silently dropped (#193) |

The first two read as confirmation for months. The third is subtler and worth reading in full on #193: the declared figure was `ON+NB+NS+PE` — the page's own recipe **minus one member** — and it looked more plausible than the correct sum would have, so nobody questioned it.

**I have added to this myself.** `IND-1800-1886`'s new declared 4,135,466 is 0.001% from its measured 4,135,440, because I derived it by subtracting measured removals from the old declared figure. It is as self-referential as what it replaced.

## What would actually check something

[...]

Issue #138, opened 2026-08-05
BLX-1850-1999 is the only dissolved federation with no successor, so nothing says Belgium-Luxembourg is Belgium plus Luxembourg


lbm364dl:
`BLX-1850-1999` (Belgium-Luxembourg, the FAOSTAT reporting union for area 15) is
the only dissolved federation in the database with **no `successor` and no
inbound `predecessor` link**, so nothing in the published data says its territory
is Belgium plus Luxembourg. Found while consuming the `successor` relation from
whep, where it is the one case that still cannot be resolved.

## Measured

Walking `successor` transitively from each dissolved federation and stopping at
the first present-day ISO3 in LUH2's 230-code vocabulary:

| polity | successors resolved |
|---|---|
| `F51-1947-1993` | 2 — `CZE SVK` |
| `F228-1945-1991` | 15 — `ARM AZE BLR EST GEO KAZ KGZ LTU LVA MDA RUS TJK TKM UKR UZB` |
| `F248-1991-1992` | 6 — `BIH HRV MKD MNE SRB SVN` (via `SCG-1992-2006`, three hops) |
| **`BLX-1850-1999`** | **0** |

The reverse direction is empty too. Every other federation has successors that
name it back in `predecessor`:

```
CZE-1993-2025  <- F51-1947-1993
SVK-1993-2025  <- F51-1947-1993
RUS-1991-2014  <- F228-1945-1991
...
BEL-1831-2025  <- NLD-1800-1830     # not BLX
LUX-1839-2025  <- NLD-1800-1830     # not BLX
```

## Why it matters downstream

whep's pre-1962 production back-cast uses LUH2 land use, which is keyed on
present-day ISO3, so a dissolved federation has no land record of its own and
its pre-1962 series is never back-cast (eduaguilera/whep#408). Reading
`successor` fixes three of the four; area 15 stays unreachable, worth 0.48% of
1961-62 FAOSTAT production tonnage. whep now warns and names it rather than
guessing, so the attribution has to come from here.

## What would resolve it

Whichever is right for this schema:

1. `BLX-1850-1999.successor = BEL-1831-2025; LUX-1839-2025` — the reading that
   makes the union's dissolution explicit and is what a consumer walking
   `successor` expects. FAOSTAT itself split area 15 into 255 Belgium and 256
   Luxembourg from 2000, matching `end_year = 1999`.
2. Or, if `successor` is reserved for state succession and a customs/reporting
[...]

Issue #48, opened 2026-07-29
ROW geometry scope and six double-claimed aggregate components need a modelling decision


lbm364dl:
Two related questions surfaced while reattributing eleven data-reporting areas out of the
FABIO Rest of World fold (#39 / eduaguilera/whep#382). Both are judgement calls about what
a residual aggregate means, so neither is patched.

## 1. ROW-1850-2023's geometry no longer matches its data scope

Eleven areas that used to fold into `ROW-1850-2023` now route to their own polities, because
they report data of their own. Eight of them are still inside ROW's polygon:

| polity | own km² | inside ROW | % |
|---|---|---|---|
| FRO-1800-2025 | 1402 | 1381 | 98.5 |
| NIU-1800-2025 | 264 | 258 | 97.6 |
| PLW-1991-2025 | 464 | 447 | 96.3 |
| BMU-1968-2025 | 72 | 69 | 96.0 |
| GLP-1816-2025 | 1582 | 1499 | 94.8 |
| REU-1946-2025 | 2623 | 2426 | 92.5 |
| MTQ-1816-2025 | 1127 | 1026 | 91.1 |
| COK-1800-2025 | 241 | 213 | 88.2 |

(Palestine, Equatorial Guinea and French Guiana were never in the union.) Total 7,319 km² of
ROW's 2,569,652.

The question: is a regional "Other" bucket a **fixed list** of territories, or a **residual**
that shrinks as territories acquire their own data? If it is a residual, the component lists
in `scripts/sources/reporting-areas/build.py` should drop these eight and the geometry should
be regenerated. If it is fixed, the overlap is intended and consumers must be told not to
allocate ROW data spatially.

## 2. Six territories are components of two aggregates each

Independent of the above, and older. Now gated by `scripts/validate_reporting_areas.py`
with these six baselined:

| component | claimed by | assessment |
|---|---|---|
| BES, CUW, SXM | `ANT-1961-2010`, `RLAM-1850-2013` | plausibly intentional succession, but the spans overlap for 1961–2010 |
| MNP, PLW | `RASI-1850-2021` (Asia Other), `ROCE-1850-2021` (Oceania Other) | looks simply wrong — both are in Oceania |
| UMI | `RNAM-1850-2021`, `ROCE-1850-2021` | genuinely straddles both regions, but the GADM polygon is one indivisible feature |

These aggregates receive data, so two of them claiming the same ground puts two aggregates'
[...]

PR #211, opened 2026-08-13
An identity for "Sudan and South Sudan, 2011 onward"

### LINKED ISSUE #139: No live entity means "Sudan and South Sudan": the post-2011 combined reporting bucket has no polity to carry
Raised from the consumer side while working eduaguilera/whep#414. Filing rather than editing,
because whep-polities is the source of truth for territorial identities and this is a proposal, not
a bug report — and because I cannot verify from here which of the two readings below this schema
intends.

## The consumer's problem

whep folds FAOSTAT areas **276 Sudan** and **277 South Sudan** into one numeric reporting bucket
(FABIO region 206) and sums them. So a whep bucket-206 value for 2015 covers **both** territories:
measured on FAOSTAT production for 2015, Sudan 53,124,088 t plus South Sudan 14,876,146 t, one
value, one bucket.

whep resolves that bucket's polity from `faostat_area_polity_map.csv`, where the rows for these
areas are:

```
area_code,year_start,year_end,polity_code,source_label,iso3,match_route,confidence,rows_observed
206,1961,2011,SUD-1956-2011,Sudan (former),SDN,manual-replace,high,139921
276,2012,2024,SDN-2011-2025,Sudan,SDN,iso-equal,high,35724
277,2012,2024,SSD-2011-2025,South Sudan,SSD,iso-equal,high,17383
```

There is no row for area 206 after 2011 — correctly, because FAOSTAT stops reporting it. whep's
year-aware lookup therefore falls back to the nearest period and labels the 2015 two-territory sum
`SUD-1956-2011`, a polity that **ended in 2011**. The extent happens to be right (unified Sudan is
Sudan plus South Sudan); the period is not, and there is no live entity in
`polities_database.csv` that means "Sudan and South Sudan, 2011 onwards":

```
SDN-2011-2025,Sudan,2011,2025,national,SDN,625,Africa,...
SSD-2011-2025,South Sudan,2011,2025,national,SSD,626,Africa,...
```

## The precedent that suggests this belongs here

The database already carries `aggregate`-typed entities for exactly this shape — a reporting unit
that is the sum of two territories, distinct from either:

```
F237-1954-1975  Vietnam (combined reporting: DRV and RVN)   1954 1975 aggregate VNM
F249-1918-1990  Yemen (combined reporting: YAR and PDR Yemen) 1918 1990 aggregate YEM
```

Both are mapped to the FAOSTAT area that reports the combination (237, 249). A post-2011
Sudan+South Sudan entity would be the same object with the time direction reversed.

## What I am not sure about, and why this is a question

For Vietnam and Yemen it is **FAOSTAT** that reports the combined figure, so the combined entity
describes a real reporting practice and belongs in a FAOSTAT-facing map. For Sudan post-2011 it is
**whep's own FABIO regionalisation** that re-sums the two successors; FAOSTAT reports them
separately. So either:

1. **This is a territorial identity and belongs here.** Add an aggregate polity for the union,
   period 2011-2025, with a composition link to `SDN-2011-2025` and `SSD-2011-2025` and a geometry
   equal to their union (which is `SUD-1956-2011`'s polygon, so no new polygon work). Then whep can
   map its bucket to it and the attribution becomes truthful. Note this entity would be live at the
   same time as both its members, which `F237`/`F249` never are — that may be exactly why it does
   **not** belong here, and I cannot tell from the schema.
2. **Or it is whep's aggregation artefact and belongs in whep**, in which case this issue should be
   closed and whep needs its own bucket-identity layer. Saying so explicitly is useful either way,
   because right now the honest answer is neither recorded nor refused.

If (1), the naming/typing questions are the same ones #47 raises for `GCT-1919-1956` (how to type a
composite reporting unit) and #48 for aggregate components, and the composition relation is the one
#138 finds missing for `BLX-1850-1999` — an aggregate whose members are not linked to it. It may be
cheapest to answer all four together.

## What whep does meanwhile

Nothing moves. whep's PR for eduaguilera/whep#414 adds a diagnostic
(`polity_bucket_coverage()`) plus a warning naming the constituent polities, so the mismatch is
loud instead of silent, and leaves the attribution exactly as it is until this is decided. Bucket
206 is the **only** bucket in whep's crosswalk whose label covers less than the bucket does;
rest-of-world 999 is already honest because its label is `ROW-1850-2023`.

Refs eduaguilera/whep#414, eduaguilera/whep#458.

lbm364dl:
> *Filed by Claude (Claude Code). Every figure measured. This answers a question rather than fixing a bug, so the evidence for the answer is the substance.*

Closes #139, which asked: does a combined-reporting entity for post-2011 Sudan belong here, or is it a consumer's aggregation artefact? It raised two doubts about putting it here. **Measuring both settles them the same way.**

## "This entity would be live at the same time as both its members, which F237/F249 never are"

`F237-1954-1975` (Vietnam, combined reporting) is live 1954-1975 **and so are both its members**, for the whole span:

```
F237-1954-1975  1954-1975  aggregate
DRV-1954-1975   1954-1975  national
RVN-1954-1975   1954-1975  national
```

The cited precedent does exactly what the issue feared — and it is not unusual: **ten more aggregates coexist with their own members**, including `RLAM`, `ROW`, `MASG`, `SYL`, `TTPI`, `BWI`, `BLI`, `LBY-1949-1951` and `IDN-1949-1969`. `LEGITIMATE_CONTAINERS` exists to record that relationship.

## "For Vietnam and Yemen it is FAOSTAT that reports the combination; for Sudan it is whep's own FABIO regionalisation"

True, and it does not distinguish the cases — because this database **already carries seven identities that exist only because FABIO folds**: `ROW`, `RAFR`, `RASI`, `REUR`, `RLAM`, `RNAM`, `ROCE`. None is a FAOSTAT reporting unit; each is a downstream regionalisation given a territorial identity here so the ground it covers can be named. A Sudan-plus-South-Sudan bucket is the same object.

So: **option (1)**.

## What it is

`F206-2011-2025`, `type: aggregate`, built as the union of the two members' **own** CShapes features (625, 626) so it cannot drift from its parts:

```
SDN-2011-2025    1,859,556 km²
SSD-2011-2025      627,258
union            2,486,805
SUD-1956-2011    2,486,812   ratio 1.0000
```

Agreement with unified Sudan to four decimal places is the check that matters: the row exists to describe exactly the territory `SUD-1956-2011` described, for the years after it stopped existing.

## `iso3` is blank, against the F-convention

`F237` carries `VNM` and `F249` carries `YEM`; this follows `SYL-1944-1953` instead. **Giving it `SDN` would assert that an ISO3 lookup for Sudan may return the two-country sum — the exact mislabelling #139 reports, moved from the period to the code.** It is reached by FAOSTAT area 206, not by an ISO3 join.

## It adds no row to `faostat_area_polity_map.csv`

FAOSTAT stops reporting area 206 in 2011 and does not resume; the post-2011 combination is FABIO's. A map row would assert a FAOSTAT practice that does not exist. whep points its own bucket at this polity instead — the division of labour #139 asks for.

## My first version declared the measured union, and A2 caught it

A declared figure copied off the geometry it is meant to test cannot disagree with it — the tautology #195 measures at **104 of 210** rows. The declared area is now **1,861,484 + 644,329 = 2,505,813**, the two *official* land areas summed, so check A has something independent to compare and the resulting **0.76%** divergence is a real measurement of how far CShapes sits from the two gazetteers. The count stays at 104 rather than 105.

## Recorded as an open question

Nothing in the schema expresses that this row **is** its two members. `predecessor`/`successor` carry chronology, not composition, so the membership is discoverable only from the page — the gap #138 finds for `BLX-1850-1999` and #48 for aggregate components.

All CI gates pass; `selftest_gates` **24/24**.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01JAxP2LSXPx67DDTWKERXiQ


lbm364dl:
**Rebased onto main after #212, and here that mattered for more than the binary conflict.**

This branch predated #212, so `git diff main..branch` listed `faostat_aliases.csv` and `registry_unmapped.csv` as changes — meaning **a merge would have reverted #212's sixteen new map rows** and put Mayotte, Curaçao and the rest back on `ROW-1850-2025`.

So: reset onto `main`, the three hand-authored files reapplied (`build.py`'s F206 builder, the `f206-2011-2025.md` page, `sud-1956-2011.md`'s successor edge — both patches clean), everything else regenerated. Verified afterwards that **both** changes coexist:

```
212 map rows present (area 270 -> Mayotte):  True
211 polity present (F206-2011-2025):         True
map mappings: 244 areas
```

All CI gates pass on the rebased head.

Issue #19, opened 2026-07-27
87% of polity pages are unreviewed drafts; ~100 assertions rest on thin ones


lbm364dl:
**645 of 740 polity pages (87%) are `wiki_status: draft`** — most written by earlier autonomous passes and never human-reviewed. `wiki/log.md` shows 33 `autonomous-*` ingests, one of which (`autonomous-38-bulk-remaining`) generated **987 pages programmatically from CSV metadata alone**, its own entry noting they have "minimal content (CSV-derived only)".

**Why this is load-bearing rather than cosmetic.** The wiki is the database's source of truth *and* the input the verification pipeline reads when judging data matches. An unreviewed page is a prior agent's hypothesis; verifying a match by citing one is circular. The pipeline now handles this explicitly — `00_intake.py` reports each page's measurable depth (bytes, source citations, `(to be documented)` markers) in the evidence bundle, and agents are told to judge on that rather than on the status label — but the underlying thinness remains.

**Current state:** 13 pages flagged `page_inadequate` by verification are listed in `state/suspect_wiki_pages.csv`. 100 assertions rest on pages under 2.6 KB with zero citations.

**Precedent that works.** `document_pages.workflow.js` documented 13 stub pages on 2026-07-24: 1.5-2.9 KB stubs became 8-12 KB pages with **145 citations, none fabricated** (`validate_citations.py` passes). Crucially it asks agents for the thing the matching pipeline needs — **territorial changes inside each row's span** — and that yielded four real findings, including the Indonesian West New Guinea gap and Brazil's unapplied Acre split.

So this is chippable: run that workflow on batches, prioritising pages that actually receive data (`state/matched_rows.parquet` gives row counts per polity).

**Note on status:** documented-by-agent is *not* `reviewed`. `reviewed` should mean a human checked the claims — 13 pages were downgraded on 2026-07-24 precisely because they claimed it without citations, and `validate_polygons.py` check D now enforces that a `reviewed` page has at least one citation. Promoting a page is a separate, human act.

lbm364dl:
## Measured, now that `wiki/index.md`'s coverage table is generated

The earlier estimates were guesses off a stale index (which claimed 1,325 polities against a real 740). Real numbers:

| metric | value |
|---|---|
| polity pages | **740** |
| `status: draft` | **655** (88%) |
| `status: reviewed` | **59** (8%) |
| pages with **zero** source citations | **188** (25%) |
| open questions (`### oq-`) | **800** |

[...]

Issue #17, opened 2026-07-27
Matchers cannot run in CI: inputs live outside the repo


lbm364dl:
The matchers cannot run in CI because their inputs live outside the repo:

| env var | default | what |
|---|---|---|
| `WHEP_LAYERB` | `~/Nextcloud/whep/layer_b/consolidated_layer_b.parquet` | the consolidated layer-B dataset (190k rows) |
| `WHEP_COMMON_NAMES` | `~/Nextcloud/WHEP_ERC 2025/.../common_names.csv` | spelling-alias table |
| `WHEP_REPO` | `/home/usuario/WHEP` | a WHEP R-package checkout, for the FAOSTAT pins cache |

Consequences:
- **Nothing validates matching in CI.** The four deterministic validators work because they need only the repo, but `01_match_and_findings.py`, `00_intake.py` and both `match.R` scripts cannot run.
- **A clean checkout cannot reproduce the pipeline**, so the 189,694-row match count is not independently verifiable.
- **FAOSTAT verification depends on the runtime environment**: without the pins cache, fixes to FAOSTAT routing cannot be confirmed and stay open (documented in `autoimprove.workflow.js`).

The data is not redistributable, so it cannot simply be committed.

**Options worth weighing:**
1. **A committed fixture** — a small, anonymisable slice (a few hundred rows spanning the tricky cases) sufficient to exercise the matchers and assert a known-good crosswalk. Cheapest path to CI-testable matching, and it pairs with the golden-test fixture in the matcher-convergence issue.
2. **Synthetic fixture** — generated rows with the same shape and no real values at all. Fully shareable; tests logic but not real-world messiness.
3. **Self-hosted runner or private artefact store** with access to the real data. Heaviest, but it would let CI verify the true match count.

Option 1 or 2 is probably enough: the validators already cover database integrity, and what is missing is protection for *matcher logic*, which a fixture can provide.

Issue #10, opened 2026-07-27
Add deterministic cross-checks over banked assertions (sums, double-counts, continuity)


lbm364dl:
Every guard built so far checks one assertion, or one polity, in isolation. Nothing checks banked assertions **against each other**, yet the errors found today were mostly relational.

**Checks worth having, all deterministic and agent-free:**

1. **Component vs aggregate sums.** Where a component and its parent both receive data for the same years and item, the component's value must not exceed the parent's. Would have caught the Ethiopia/AOI fold-up and the Manchuria-into-China case directly.
2. **Double-count detection.** If an aggregate polity and its constituents both carry the same item and years, summing them double-counts. Flag the overlap — the Djibouti-coffee entrepôt finding is this shape (coffee attributed to both French Somaliland and Ethiopia).
3. **Magnitude continuity at banked boundaries.** Where an assertion is banked `verified_equal` (territory asserted constant), a large step in the series across that boundary contradicts the verdict. A cheap way to find confident-wrong confirms without re-running agents.
4. **Per-km² plausibility across the whole DB.** `05_magnitude_screen.py` already does a version of this; the missing piece is the *relational* comparison of a polity against its own family rather than against the global item median.

The value: these find contradictions **between** verdicts without trusting any single one, which is the one gap that blind review cannot close (two agents can be wrong the same way, but they cannot both be right if their conclusions contradict each other arithmetically).

`05_magnitude_screen.py` and `06_landuse_consistency.py` are working precedents — the land-use check both detects bad cells *and recovers the correct value*, which is the standard to aim for.

PR #272, opened 2026-08-17
Triage the 796-assertion backlog: order it, bound it, and settle the 20 pairs arithmetic can settle (#7)

### LINKED ISSUE #7: Verification backlog: 723 pending assertions
`state/assertions.json` currently has **723 pending** assertions (of 1,031). An assertion is one `(label, source, year_segment) → polity` claim; 190k layer-B rows collapse into ~1,031 of them, which is what makes exhaustive verification affordable.

**Process** (see `pipelines/polity-autoimprove/README.md`):

```bash
python3 pipelines/polity-autoimprove/00_intake.py --input <parquet> --label-col country --year-col year \
  --iso-col iso3c --value-col value --item-col item --unit-col unit --period-col period \
  --aggregate-col is_aggregate --prior-code-col polity_code --source-col source
# take pending keys from state/assertions.json, chunk ~50
Workflow verify_assertions.workflow.js  args {review_sample: 5, out: "verdicts_chunkN.json", keys: [...]}
# inspect state/verdicts_chunkN.json, then
python3 pipelines/polity-autoimprove/apply_verdicts.py verdicts_chunkN.json
```

**Track record so far** — 150 banked across four chunks, 0 quarantined on routing, but the verdicts are not rubber stamps: they cite magnitudes and prior wiki decisions, and 13 of 50 in one chunk self-flagged `best_available`. The pass has surfaced, among other things, the Ethiopia/AOI fold-up, the Djibouti coffee entrepôt, the Brazil unapplied split, and three page-vs-database contradictions.

**Cost** roughly 35k Sonnet tokens per assertion including reviews, so ~25M for the remainder. Chunks are independent — this can be chipped away at.

Worth ordering chunks by rows descending (biggest data impact first); the tail is small-row assertions where evidence is thinnest.

lbm364dl:
> *Filed by Claude (Claude Code). Every figure measured against the repo.*

Issue #7 asks for a triage of the pending assertion queue: verify what deterministic evidence can settle, and say what genuinely needs a human. This adds `pipelines/polity-autoimprove/12_triage_assertions.py` (spends no tokens, writes no verdict) plus two committed snapshots of its output, since `assertions.json` is gitignored and the queue is otherwise invisible to anyone reading the repo.

## The issue's headline number is stale

| | issue #7 | measured 2026-08-17 on current main |
|---|---:|---:|
| assertions | 1,031 | **1,073** |
| pending | 723 | **778 pending + 18 reopened = 796** |
| rows behind the backlog | — | **94,925** of 189,839 routed |

The queue **grew**: polities and aliases were added over the last two days, and banked assertions reopen when their evidence hash changes. Fresh `00_intake.py` run against `consolidated_layer_b.parquet`.

## What deterministic evidence cannot do — and why that is structural, not a gap

An assertion exists *because* the deterministic pass already finished: `matchlib` decided the route (alias → iso → name, with year containment) and recorded it. What is pending is whether the **source's** reporting territory under that label equals the polity's territory over that segment — and the repo's tables are one side of that comparison, so no check over them can close it.

**659 of the 796 carry no deterministic signal at all.** Those genuinely need the reading pass. The script writes no verdict and never touches the review ledger.

## What it can do: source-internal nesting, then the panel's own arithmetic

Every unit screened is extensive (tonnes, ha, heads…), so if a label's figures included a nested territory the *same source* reports separately, `outer >= inner` must hold in every shared (item, unit, year) cell.

* **234** same-source nesting pairs (44 outer / 66 inner polities), touching 189 assertions, 137 pending.
* In **20** of them inclusion is arithmetically **impossible** → the source reports the outer territory *exclusive* of the inner, so the polity's **territory basis**, not the routing, is wrong.
* **13 of the 20 involve an already-banked assertion**, so the screen contradicts recorded verdicts: re-verification, not first verification.

The two largest (all re-derivable from `state/assertion_nesting_flags.csv`):

* IIA's and Mitchell's `japan` → `JPN-1895-1945` (626,507 km² — metropolitan Japan + Korea + Formosa) while both sources report Korea and Formosa as separate labels throughout. Japan below Korea in **220 of 394** shared IIA cells, below Formosa in **120 of 330**. Those series are metropolitan Japan.
* Juan's `united kingdom` → `GBR-1800-1921` (313,550 km² — Great Britain **and** Ireland) while Juan reports `ireland` separately. 1901 flax fibre area: Ireland **22,436 ha** vs the UK's **259 ha**; UK below Ireland in **87 of 637** cells, concentrated in potatoes, flax and linseed. Juan's UK is Great Britain.

Also flagged: FAO-1952's 1938 `germany` below its own `germany western`/`germany eastern`; `indochina viet nam` below Cambodia and Laos; IIA's `india` below Burma; Mali below Upper Volta in IIA and Mitchell.

The nesting flag (not the arithmetic) independently re-finds the **Ethiopia/AOI fold-up** the reading pass found by hand — Mitchell's `ethiopia` → `AOI-1936-1941` (1,712,684 km²) against its own `somalia` → `ITS-1908-1960` (464,286 km², wholly inside it). That pair shares no comparable cell, so it is calibration, not a second settlement; the README says so explicitly rather than implying the arithmetic caught it.

**Consistency is not proof**: the 55 pairs where `outer >= inner` holds everywhere look exactly like a genuine double count would, and 118 pairs share no comparable cell. Those stay for the historian.

## The ordering the budget should follow

`state/assertion_triage.csv` gives every pending/reopened assertion a `verify_order` and a first-match tier:

| tier | assertions | rows |
|---|---:|---:|
| `territory_basis_wrong` | 17 | 3,184 |
| `nested_reporting` | 120 | 5,563 |
| `boundary_year` | 1 | 10 |
| `weak_route` | 122 | 12,535 |
| `precedent` | 256 | 36,495 |
| `thin` | 53 | 272 |
| `bulk` | 227 | 36,866 |

Two bounds worth knowing before buying ~25M tokens: only **66 of 796** candidates have a `reviewed` wiki page, so for the other 92% a confirm resting on the page alone is recorded as `best_available` by `apply_verdicts.py`'s anti-circularity rule (see #19) — the pass buys routing assurance, not territorial proof. And of the 159 verdicts applied so far only **3 (1.9%)** changed a routing.

The inclusion test is also the first concrete instance of the deterministic cross-checks #10 asks for, applied to the queue rather than to one polity.

## Left open, deliberately

* **No verdicts applied**; no polity, alias or wiki page changed. The backlog is ordered and bounded, not banked — so this does **not** close #7.
* Fixing the 20 settled pairs needs a target that does not exist today: there is no metropolitan-Japan row for 1895-1945 and no Great-Britain-without-Ireland row for 1800-1921. Choosing between creating those rows, re-routing to a smaller existing row, or recording a source convention over an over-scope proxy is a historian's call, filed separately rather than guessed here.
* Not wired as a CI gate: its input derives from a parquet outside the repo, so CI cannot recompute it (#17). Re-run after `00_intake.py`.

## Verification

* All 57 gate scripts in `.github/workflows/validate.yml` run; **all pass except one pre-existing, environment-only failure** (below).
* `python3 scripts/selftest_gates.py` → **PASS: 40 gate(s) fail on an injected defect and name it, and every gate runs in CI**. No new gate added, so no new CASES entry is required.
* No generated file changed: this PR touches no wiki page, no polity row and no polygon, so `build_database.py --check`, `write_manifest.py --check`, `update_wiki_index.py --check`, `04_territory_basis.py --check` and the label/alias/faostat/iso3 `--check` modes all pass unchanged.

**Pre-existing failure, not caused by this branch:** `python3 scripts/write_feature_index.py --check` reports `polygon_feature_index.csv` stale. Regenerating changes exactly one line — `reporting-areas,RASI-1850-2025,3,,,2127.9` → `1139.6` — i.e. the *shared* `data/geodata/reporting-areas/reporting_areas.gpkg` no longer produces the areas main's committed index was built from. `data/final/` here is byte-identical to main, so the mismatch is between main and the shared geodata directory, not this change; the regeneration was reverted rather than committed so this PR cannot bake a corrupted area in. Worth checking on main before the next merge.

Refs #7 (triage delivered; the 796 verifications themselves remain), #10.


lbm364dl:
The territory-basis half is filed as #273 (the 20 arithmetically-impossible pairs, with the three options each needs).

Issue #360, opened 2026-08-18
Where layer B splices a series across sources, 30% of the joins jump by over 30% — and 30 jump by over 100x


lbm364dl:
*Filed by Claude, found while trying to cross-check the wine factor in #358 against another source.*

Layer B assigns each cell to exactly one source, and where a series switches source mid-stream the
join is usually **not scale-consistent**. Measured across the whole panel:

    series (country, item, unit)                    13,430
      containing more than one source                 574   (4%)
    adjacent-year source splices                    1,245
      where the value jumps by more than 30%           373   (30%)
      by more than 3x                                  123   (10%)
      by more than 10x                                  66   (5%)
      by more than 100x                                 30   (2%)

A jump at exactly the year the source changes is not a harvest, a war or a border. The worst are
physically impossible:

    united states  flax fibre and tow  ha   1949->1950  juan -> mitchell   x1655
    united states  flax fibre and tow  ha   1951->1952  juan -> mitchell   x1337
    canada         flax fibre and tow  ha   1956->1957  juan -> mitchell   x1231
    bulgaria       rapeseed            ha   1916->1917  iia  -> juan       x90.9
    turkey         wheat               ha   1940->1941  iia  -> mitchell   x41.1

And it reaches staples, where it matters most:

    turkey     wheat       ha      1940->1941  iia->mitchell     x41.07
    uruguay    barley      tonnes  1903->1904  juan->mitchell     x9.11
    paraguay   wheat       ha      1956->1957  juan->mitchell     x7.50
    bolivia    wheat       tonnes  1928->1929  mitchell->juan     x6.92
    ecuador    rice paddy  ha      1932->1933  juan->mitchell     x2.92

### It is not one bad source

Splices with a >3x jump are near-symmetric across every pairing — `iia->mitchell` 26,
`mitchell->iia` 25, `juan->iia` 20, `iia->juan` 19, `juan->mitchell` 19, `mitchell->juan` 14. So this
is a property of the splicing, not of one source being wrong.

### What I am not claiming

I do not know the cause per splice. Candidates, and they are probably all present: a unit or
thousands-multiplier mismatch (the 1000x+ flax cases look like exactly that), genuinely different
territorial coverage between the two sources' labels, different item definitions (seed cotton vs lint,
fibre vs tow), or a source-specific reporting basis. Distinguishing them needs the sources, one splice
at a time. The measurement here is the jump and the year it lands on.

[...]

PR #378, opened 2026-08-18
An iia item series can switch RAW PRODUCT year to year

lbm364dl:
28 layer-B `iia` item series are stitched from several **different raw products**, switching year to year — so their level shifts are product switches, not changes in output.

## The exhibit: australia / sugar raw centrifugal / tonnes

```
1912  131,961   sugar: cane
1913      922   sugar: BEET             x0.01
1919    1,576   sugar: beet
1920  165,616   sugar: cane             x105
1921  297,700   sugar: cane
1922    2,829   sugar: BEET             x0.01
1925  537,212   sugar: cane, unrefined  x175
1938  783,200   sugar: cane
1939    6,400   sugar: BEET             x0.01
1944  611,300   sugar: cane             x873
```

Australia's beet sugar industry was negligible, so **12 of those 33 cells are about 1/300 of the series they sit in**.

**The switches go back and forth**, which is what rules out growth as the explanation — a monotonic series drawing on an early product and a late one would show the same level gap innocently.

## Why nothing here could see it

- The **source never changes**, so `16_source_splices.py` finds no seam.
- The values are entirely plausible **for the product they came from**, so no magnitude screen fires.
- The **item name is constant**, so nothing keyed on the item notices.

It's also the mechanism behind part of #360's level-shift confound: a detector looking for territorial steps in these series finds the yearbook's choice of table instead.

## This refutes part of a convention I registered myself in #369

That's the more useful half of this change. The entry said IIA's `p`/`n`/`k` are the fertilizer macronutrients.

**What stands:** they are about fertilizer nutrients, and must not be concatenated with `fao1952`'s `commercial *` series.

**What is wrong:** the implication that such a series is a **nutrient total**. Traced per cell:

```
sweden p:   1909-1920  calcium superphosphate   102,509 -> 185,320
            1930-1932  BASIC SLAG                14,520 / 10,171 / 7,730    x0.08
            1933       calcium superphosphate   219,030                     x28
japan  n:   calcium cyanamide throughout, except 1919 and 1933 (ammonium sulfate, x27 and x5)
```

Superphosphate, basic slag, natural phosphate rock, ammonium sulfate and calcium cyanamide are **different materials with different nutrient contents**. So a `p` series is a patchwork of materials rather than P₂O₅, and no trend should be read from one. Also affected: norway, austria, belgium, france, united kingdom, canada, united states, japan.

The convention rows now carry the refutation plus a re-test note; `check_iia_npk` tests only the parts that stand, and still passes.

## Other affected series

`czech republic / grapes / ha` draws on **five** products in a 17-value series — `grapes: table` (8 ha), `grapes: wine` (17,298), `vineyards: all` (27,000), `vineyards: grapes, raisins` (2,000). Plus `congo`/`niger` cotton lint (`cotton` vs `cotton: ginned`, 8–11×), grapes for australia/cyprus/japan/israel/turkey, raisins for usa/australia, and the already-logged `wheat` = spelt/meslin.

## One methodological caveat, recorded rather than buried

The first version of this measurement compared each product's **median without controlling for year**, which would report a level gap for any growing series that happens to draw on an early product and a late one. That's why the finding rests on the **oscillation**, traced year by year — not on the median comparison that surfaced it. 28 is the count that survives; I'm not claiming all of them are as clean as Australia's sugar.

## Verification

60 gates pass; conventions re-test clean (18 conventions, 0 contradicted).

Created by Claude.


lbm364dl:
*Comment by Claude.*

Now gated, in #379 — and the count went **28 → 9**, which is a correction to this issue rather than a refinement of it.

This issue measured 28 series by comparing each product's **median level**. That test is confounded: a *growing* series which happens to draw on an early product and a late one shows the same level gap as a real defect while being perfectly innocent. I flagged the confound here but still reported the 28.

The gated criterion is **oscillation** — a product recurring after another has intervened (A…B…A), traced year by year, which no trend can produce. **9** series survive:

| jump | series | switches between |
|---|---|---|
| **185×** | australia / sugar raw centrifugal | `sugar: cane` ↔ **`sugar: beet`** |
| 72× | czech republic / wheat | `meslin` ↔ `spelt` |
| 47× | japan / p | natural phosphate ↔ superphosphate |
| 28× | sweden / p | basic slag ↔ superphosphate |
| 27× | japan / n | calcium cyanamide ↔ ammonium sulfate |
| 18× | france / p | 3 products |
| 14× | spain / sugar raw centrifugal | beet ↔ cane |
| 10× | austria / p | superphosphate ↔ other |
| 4× | paraguay / sugar raw centrifugal | cane, unrefined ↔ cane |

Every headline case in this issue is among the 9, so the stricter test **kept the evidence and dropped the inference**. The 19 that fell away were not shown to be innocent — they were shown to be unproven, which is a different claim and the honest one.

Two other things the gating exposed:

- `australia / sugar` 1942 reads **700**, a value that also appears in the raw extract under castorseed, coffee and butter. Cells like that are now counted **ambiguous** rather than attributed, and `n_ambiguous` is published per row, because a value shared by four products is not evidence about any of them.
- `czech republic / grapes`, which this issue described as drawing on five products, does **not** appear in the gated 9. Its cells are largely ambiguous or single-product-per-value, so the five-product claim rests on matches too thin to gate. The series is still suspect — it is the one with 8 ha beside 27,000 ha — but the count of five is not something I can defend.

The gate restates `MIN_CELLS`, `MIN_PER_PRODUCT` and `LEVEL` rather than importing them, so relaxing the generator and regenerating cannot quietly refill the table; the selftest injects exactly that shape. Selftest 71 → 72 gates.

lbm364dl:
*Comment by Claude.* Independent corroboration of the `grapes` scope mixture, from a direction that was not looking for it.

I was testing derived-product ratios across the panel — oil from olives, wine from grapes, sugar from cane, butter from milk — to see whether the flax fibre/linseed conflation had siblings. Most of those tests turned out to have no power (the ratio is an economic choice, not a physical constant, so there is no defensible band). But one result is impossible rather than merely wide:

```
wine / grapes, tonnes:   115 shared cells   median ratio 6.024
```

Wine cannot weigh **six times** the grapes it came from. The physical figure is around 0.7.

That is explained entirely by this issue's finding: layer B's `grapes` item alternates between `grapes: table` — a small subset — and `vineyards: all`, so for the years where it holds the table-grape figure, the denominator is a fraction of the real crop. The worst cells make it plain: `iia germany 1937` reads a ratio of **325.6**, and `iia greece 1933` **234.4**.

So the mixture recorded here has a second, independent signature: **it breaks the wine-to-grape identity**, on 88% of the 115 cells where both items appear.

Two caveats on how much weight that carries:

- It is corroboration, not new evidence. The mixture was already established per-cell against the raw extract, and `czech republic / grapes / ha` is in the gated table with its five products.
- The ratio test itself is not a reliable detector, which is why I am not proposing one. Of the four derived pairs I tried, three have bands too soft to distinguish a defect from a different farm economy — olive oil/olives varies with the table-olive share, butter/milk with how much milk is sold liquid. The wine case is only usable because the observed ratio is *physically impossible*, not merely unexpected.

Issue #451, opened 2026-08-19
1,995 collapse groups publish a blended value: 1,577 true duplicate keys and 418 item-code collisions, and only 3.0% are reachable by a label-pair screen


lbm364dl:
*Filed by Claude. Systematic sweep generalising #411 (Germany) and #449 (China).*

Both of those issues describe one polity receiving a label **and** a longer label extending it. I swept for the
pattern across the whole panel and measured cell-level overlap, which is what separates a real collision from a
harmless coincidence of naming.

```
label pairs where one label extends another on the SAME polity      46   across 22 polities
   ...of which share at least one (item, unit, year) cell           12   across  9 polities
   ...of which the shared values DIFFER                              9
   ...of which the shared values are IDENTICAL                      3
```

**34 of the 46 never share a cell** — `Zanzibar`/`Zanzibar and Pemba`, `United States`/`United States and
dependent territories`, `rwanda`/`rwanda and burundi` and others are complementary in time or item, so they
carry no double-count risk. That is worth stating because the structural pattern looks alarming and mostly is
not.

## The 12 that do overlap

```
polity             whole                     part                            shared  ident  differ
DEU-1920-1938      Germany                   Germany Western                     22      0      22   #411
DEU-1920-1938      Germany                   Germany Berlin                       8      0       8   #411
GBR-1921-2025      United Kingdom            United Kingdom Great Britain         6      0       6   NEW
ETH-1907-1936      ethiopia                  ethiopia pdr                         4      4       0   NEW
ETH-1941-1952      ethiopia                  ethiopia pdr                         4      4       0   NEW
KOR-1948-2025      Korea                     Korea South                          4      2       2   NEW
CHN-1947-1949      China                     China 22 provinces                   3      0       3   #449
CHN-1947-1949      China                     China Manchuria                      3      0       3   #449
CHN-1949-1950      China                     China 22 provinces                   2      0       2   #449
DEU-1920-1938      Germany                   Germany  Western                     1      0       1   #411
ETH-1936-1941      ethiopia                  ethiopia pdr                         1      1       0   NEW
JAM-1800-2025      BWI Jamaica               BWI Jamaica and Dependencies         1      0       1   NEW
```

The sweep confirms #411 and #449 are the two largest instances (31 and 8 cells), which is reassuring about how
they were found. Four groups are new.

## KOR-1948-2025 is the serious one: the whole peninsula is on South Korea's polity

```
item                              year   Korea      Korea South
[...]

Issue #411, opened 2026-08-19
DEU-1920-1938 holds the 1937 Reich total AND its post-war subdivisions across 22 items, so the published figure is a median 74% of the Reich


lbm364dl:
> ### ⚠️ Corrected — see [this comment](https://github.com/eduaguilera/whep-polities/issues/411#issuecomment-5340599826)
>
> **The "58.2%" below is wrong and overstates the harm. The right figure is a median 74% per cell.**
>
> I computed it as a ratio of two **sums** taken across all 22 colliding cells — and those cells carry four different units (tonnes, hectares, heads, people). Adding those together is meaningless, and the ratio is then dominated by the largest cells, which happen to have the lowest individual ratios. Per cell: median **73.9%**, mean 70.0%, range 50.6%–85.7%.
>
> The mechanism, the 22-cell count, the direction, and the per-item table below (rye 53%, oats 55%, milk 64%) are all unaffected — those were computed per cell and are correct as printed. Only the summary statistic was wrong: it should have been a median of ratios, never a ratio of sums.

*Filed by Claude, found while building the same-polity overlap census for #355 (PR #407).*

`DEU-1920-1938` receives **four** fao1952 labels, all at 1937:

```
Germany            41 rows      the actual 1937 Reich
Germany Western    36 rows      the post-war West German area's 1937 output
Germany Berlin     14 rows      Berlin's
Germany  Western    3 rows      a doubled-space spelling of the second
```

`Germany` and `Germany Western` collide on **22 (item, unit, year) cells**, and `Germany Berlin` on 8 more. This is the largest overlap in the census.

## The labels are not an extraction defect — the ratios prove it

`Germany Western` over `Germany`, per item:

```
0.400  dry peas [t]           0.641  oats [t]
0.409  dry peas [ha]          0.652  potash fertilizers
0.548  rapeseed [t]           0.655  oats [ha]
0.561  rapeseed [ha]          0.669  population
0.579  rye [ha]               0.684  linseed [t]
0.598  rye [t]                0.690  linseed [ha]
0.600  dry beans [ha]         0.693  phosphoric acid
0.612  nitrogenous fert.      0.714  dry beans [t]
0.632  mixed grain [ha]       0.729  poultry
0.638  mixed grain [t]        0.736  butter
                              0.749  milk
```

[...]

Issue #367, opened 2026-08-18
3,081 fao1952 rows are indistinguishable within the panel, and in 239 groups a TOTAL sits beside its own parts — summing them double-counts exactly


lbm364dl:
*Filed by Claude. Found while building a single-year spike detector: its `fao1952` hits kept listing a "neighbour" in the same year, which turned out not to be a bug in the detector but two rows for one series-year.*

## The finding

1,130 `(country, item, unit, source, indicator, year)` groups in layer B carry **more than one row** — 3,081 rows in total, all `fao1952`. And **no column in the panel distinguishes them**: `item_code`, `source_detail`, `period`, `polity_code`, `iso3c` and `indicator` are identical across every member.

```
indistinguishable (series, year) groups     1,130     rows  3,081
  group sizes    2 -> 533    3 -> 380    4 -> 211    5 -> 5    6 -> 1
  all fao1952; 30 distinct items
```

The cause is visible in the item labels: they are **collapsed multi-commodity ranges**. `horses mules asses` is three species. `centrifugal sugar noncentrifugal cane sugar` is two commodities. `meat`, `other meat`, `poultry`, `milk` likewise. Afghanistan 1949 `horses mules asses` (1000 heads) reads 1,000 / 500 / 200 — three species under one label, and `item_code` is the *range* `110_117`, so which row is which animal is not recoverable from the panel.

## The serious half: a total indistinguishable from its own parts

In **239 of the 1,130 groups (21%)**, the largest value equals the sum of the rest to within 2%:

| country | item | unit | year | total | == parts |
|---|---|---|---|---|---|
| united states | meat | 1000 t | 1950 | **10,015** | 4,884 + 4,860 + 271 = 10,015 |
| united states | meat | 1000 t | 1951 | 9,938 | 5,209 + 4,492 + 237 = 9,938 |
| italy | grapes | 1000 t | 1951 | 7,318 | 4,435 + 2,883 = 7,318 |
| united states | centrifugal + noncentrifugal sugar | 1000 t | 1950 | 2,337 | 1,825 + 512 = 2,337 |
| france | meat | 1000 t | 1949 | 1,805 | 965 + 770 + 70 = 1,805 |
| argentina | meat | 1000 t | 1950 | 1,184 | 990 + 125 + 69 = 1,184 |

[...]

PR #457, opened 2026-08-19
Screen the key the consumer actually collapses on: 1,977 groups publish a blended value, only 3.0% reachable by a label-pair screen (#451)

lbm364dl:
Addresses #451; the root is #367.

The WHEP R package reduces duplicate keys with `mean(value, na.rm = TRUE)` over
`(year, area, item…, unit)` — **no label column, no source column**. Nothing screened that key.

### Why `25_same_polity_overlaps.py` was not already this

That screen reads the same panel but groups by `(polity, source)` and reports label **pairs**, so two
classes of collapse are outside what it can see by construction: duplicates under a **single label**, and
duplicates spanning **several sources**. Measured, those are nearly all of it:

```
collapse groups holding >1 row                        2,345   (5,762 rows)
  values identical                                      368   mean returns the value unchanged
  values DIFFER                                       1,977   published value is a blend of all members
     one label only                                   1,436
     more than one label anywhere                       541
     a SINGLE source contributing two labels             60    = 3.0%  <- reachable by a pair screen
```

`germany` alone supplies **ten** rows for population 1937 on `DEU-1920-1938`, so there is no pair to
report. `same_polity_overlaps.csv` holds 24 rows, consistent with that ceiling. **The label-pair lens is
not a partial view of this defect; it is very nearly disjoint from it.** I first reported the reachable
share as 27% on #451, reading it off the label count — the per-source grouping makes it 3.0%, and the
comment there is corrected.

### What the differing groups publish

The framing in #411/#449/#451 was that a polity wrongly carries the *larger territory's* figure. It does
not — it carries an average of both, a number from no territory at all:

```
KOR-1948-2025 population 1951    peninsula 29,300 | South Korea 20,500  -> publishes 24,900
DEU-1920-1938 population 1937    Reich 57,576, ten rows in the group    -> publishes 13,854.5  (24%)
CHN-1949-1950 pigs 1949          china 95 | 22 provinces 59,510         -> publishes 29,802
```

The dominant items are livestock and meat aggregates — `horses mules asses` 245, `meat` 203, `poultry`
140 — not the population table (12.4%), and for 4-row groups the ratio clusters at **0.25**, the signature
of one total beside three small components. That is #367's `total_beside_parts`, counted polity-keyed
across all five sources (1,977) rather than fao1952-only (1,130).

### What is pinned, and what deliberately is not

Screening counts are **printed, not pinned** — they move on every honest regeneration, and a gate that
pinned them would fail on legitimate work. Pinned both ways: three **curated anchors** quoted in open
issues, plus arithmetic that cannot legitimately break — a stored mean outside its own group's range is
describing a `sum`, not a `mean`.

### Two things the build caught rather than assumed

- **The ratio disagreed with the table's own values** on 4 rows. Nyasaland `other meat` 1949 stores mean
  `0.233333` over v_max `0.3`, which is `0.777777`, while the unrounded ratio rounds to `0.777778`. Fixed
  in the writer by deriving the ratio from the values **as written**, so each row is exactly
  self-consistent and the gate can recompute with a tight tolerance. The gate found this on its first run.
- **The arithmetic mutation first tested two arms at once.** It picks the largest group by row count —
  which *is* a curated anchor (DEU, ten rows) — so it tripped check C as well and would have passed with
  the arithmetic arm deleted entirely. It now reads `ANCHORS` from the gate and excludes them; verified
  exactly one problem, from arm B.

### Corrections to my own claims on #451, carried into the code

- **`ethiopia pdr` does not double-count anything.** All 9 shared cells are byte-identical and `mean`
  returns the value, so the duplicate is cosmetic under the current consumer. Recorded as an anchor
  precisely so the withdrawal stays honest — if it ever reads `values_differ`, it became a real fix.
- **`germany`/`germany western` share ONE cell, not 22.** The 22 counts *items on the polity* in #411.
- `25_same_polity_overlaps.py`'s docstring **already** recorded the mean-not-sum fact from #367. The issue
  text contradicted the repo's own tooling.

Gates 71 → 72, selftest 91 → 93. All 72 gates pass; full 93-case selftest green.


lbm364dl:
*Comment by Claude.* Pushed a fix to a defect in this PR itself: **the generator was numbered `27_`, one
filename apart from the existing `27_series_collapses.py`, and both say "collapse" for different things.**

- `27_series_collapses.py` — a **series** dropping to zero and staying there. A level break, over time,
  within one series.
- this tool — the single-year **aggregation** the consumer performs when several rows share its key.
  Nothing drops and nothing is about time.

Same number plus a near-identical name is a reading trap, and I walked into it by choosing the next free
number without checking what the neighbouring one meant. Renamed to
`30_consumer_key_collapse.py` (28 and 29 are taken by `28_item_blocks.py` and `29_era_shift_verdicts.py` in
#459), and both files now state the distinction explicitly so the next reader does not have to infer it.

Found while checking whether an unrelated measurement was already in tracked state — which it was
(`impossible_pairs.csv` already holds the 176 zero-area rows I had just re-derived, so no new issue there
either).

No behaviour changes: the table is byte-identical, `--check` reports current, the gate passes, and both
selftest cases still bite.

Issue #434, opened 2026-08-19
The triage queue was stale because its generator CRASHED on null spans (fixed in #458); the resync itself is blocked on #310's period-year rule, and re-keys 291 of 1,074 assertions


lbm364dl:
*Filed by Claude.*

While wiring the last of #432's three ungated tables I checked `assertion_triage.csv` against
`assertions.json` and they do not describe the same set of assertions. The queue was built against an
earlier assertion set whose year spans have since moved, and nothing re-syncs them.

## The evidence

```
queue rows (assertion_triage.csv)                    796
assertions.json                                     1073   (917 pending/reopened)

queue rows whose key no longer exists in assertions.json     71    (iia 44, fao1952 27)
pending/reopened assertions absent from the queue           198    (fao1952 82, iia 75,
                                                                    mitchell 32, juan 9)
```

It is **re-spanning**, not a status filter — I checked that first and my initial reading was wrong.
For all **71 of 71** orphaned queue rows the same `(label, source)` is present in `assertions.json`
under a *different* span, and none is absent entirely:

```
queue 1929-1941   vs assertions ['1933-1941']    albania|iia
queue 1938-1951   vs assertions ['1949-1951']    anglo egypt sudan|fao1952
queue 1929-1945   vs assertions ['1930-1945']    benin|iia
queue 1938-1951   vs assertions ['1949-1951']    british north borneo|fao1952
```

This is the same mechanism `verdict_carryover.csv` exists to absorb: a span moves, and everything
keyed on the old span is left pointing at nothing.

## Why it matters

The triage queue decides **what gets looked at**, so drift here does not produce a wrong number — it
produces work never done, which is invisible by construction.

- **8,533 panel rows (8.9% of the pending pool)** sit behind the 198 assertions that never reach the
  queue.
- **4 of those 198 are arithmetically-impossible inclusions** — `#273`'s class, where the source
  reports the outer territory exclusive of the inner. They are condemned in
  `assertion_nesting_flags.csv`, are neither queued nor adjudicated, and so nothing will ever look at
  them: `germany berlin|fao1952|1938-1939`, `germany western|fao1952|1938-1948`,
  `indochina laos|fao1952|1936-1951`, `ryukyu islands|fao1952|1937-1939`.
- The 71 dead queue rows are **not** concentrated in low-scrutiny tiers: 3 are
  `territory_basis_wrong` and 16 `nested_reporting`, the two tiers that exist because those rows need
[...]

Issue #310, opened 2026-08-18
9,210 valued rows with no year are assigned to year-bounded polities, half of them where several spans were possible


lbm364dl:
*Filed by Claude, found while corroborating a source convention in the third verification pass.*

9,865 rows in the consolidated layer-B panel have **no year**. Every one of them carries a
value, and 9,210 of them are assigned to a polity whose whole identity is a year range.

    layer B                        192,670 rows
      null year                      9,865   (5.1%)   all carry a value
        iia                          6,163   (23.5% of iia)
        fao1952                      3,702   (15.6% of fao1952)
        juan / mitchell / sa_colonial    0

    reach matched_rows.parquet       9,334
      ASSIGNED to a polity           9,210

The concentration is the first thing worth noting: Juan, Mitchell and sa_colonial have zero
null years between them. This is not a general panel weakness, it is specific to the two
yearbook sources, which is where a reader would expect the year to be least ambiguous.

### Why assignment is the problem, not the null itself

A polity code IS a year range. `JPN-1895-1945` asserts fifty specific years; `BGR-1919-1940`
asserts twenty-two. A row with no year cannot be known to fall inside either. Yet:

    JPN-1895-1945    192 null-year rows
    BGR-1919-1940    163
    USA-1867-1959    215
    ESP-1800-2025    203

**4,605 of the 9,210 (50%) were assigned to an ISO3 that has more than one candidate span.**
Romania has eight spans and its null-year rows landed split across ROU-1913-1918 and
ROU-1920-1940; Greece has eight and split across GRC-1881-1913 and GRC-1919-1947. Something is
discriminating between spans for rows that carry no year, and whatever that signal is, it is
not the row's own date.

The other 4,605 went to an ISO3 with a single span, where the assignment is at least not a
choice between alternatives — worth separating, because those are defensible and the first
group is not.

### It has already reached the verification queue

13 assertions carry a literal null span, routing to tightly bounded candidates:

    turkey|iia|None-None          -> TUR-1913-1914     (a two-year window)
    china mainland|iia|None-None  -> CHN-1913-1914
    poland|iia|None-None          -> POL-1815-1918
    nigeria|iia|None-None         -> NGA-1886-1914

An assertion with no years cannot be verified against a candidate whose entire claim is a
two-year window — there is nothing for a verifier to check. They total only 70 rows, so
exposure-ordered selection will never reach them and they will sit in the queue indefinitely.

[...]

Issue #308, opened 2026-08-18
Re-spanning orphans banked verdicts (15.7% of 466 assertion-level, all span drift) and the row figure depends on which record is the authority: 9,481 against a regenerated queue, 38,056 against the tracked one


lbm364dl:
*Filed by Claude while selecting the third verification tranche.*

Re-deriving the assertion queue renames its keys, and 51 of the 228 banked verdicts —
22% of every verification decision made so far — no longer match anything in it.

    banked verdicts                                  228  (246 applied records)
      still matched to a queue entry                 177
      ORPHANED                                        51
        same label|source, different span             51
        label|source gone entirely                     0

All 51 are span drift, not deletion. `algeria|iia|1920-1945` was verified; the queue now
carries `algeria|iia|1919-1945`. `brazil|mitchell|1879-1960` was verified; the queue now
carries `1909-1960`, `1879-1902` and `1903-1908`. The evidence did not change and the
judgement did not become wrong — the label the judgement was filed under did.

**34,143 rows currently sit in the queue under a re-spanned key.** They read `pending`,
so the next tranche selects them and pays to decide them again.

### The queue's own status field is not the state either

Independently of the above, 49 assertions that DO match a banked verdict still read
`pending` (7) or `reopened` (42). `apply_verdicts.py` writes the ledger; only a re-run of
`00_intake.py` re-labels the queue. #307 works around this by treating
`verdicts_applied.jsonl` as the authority, which is correct for selection but does not
recover the 51 — nothing in the queue points at them.

### What the India case shows about the blast radius

`india|mitchell|1915-1937` was quarantined in the first pass: the reviewer found a −18.6%
one-step break at 1928/29. Intake then re-split India into seven spans, and the new
`india|mitchell|1914-1936` (595 rows) **still contains 1928/29**. The re-derivation did not
act on the finding, because it could not see it.

That one was caught anyway, and it is worth being precise about why: `quarantine.csv` is
tracked and is keyed independently, so the tranche-3 verifier found the prior work there and
cited it. It then improved on it — the break is specific to **rice, paddy** (1927=31,755,000
ha, 1928=33,700,000, 1929=27,421,000; wheat/barley/maize smooth), i.e. Mitchell drops Burma
from the reporting territory at 1929, eight years before the 1937 political separation.

So quarantines have a carrier. **Confirms do not.** 48 of the 51 orphans are ordinary banked
verdicts with no tracked artifact keyed to find them by, and those are simply redone.

[...]

PR #458, opened 2026-08-19
The triage queue is stale because its generator crashes: fix the null-span crash, and the resync turns out to be blocked on #310 (#434)

lbm364dl:
Addresses #434.

`span()` did `int(x["years_observed"].split("-"))`, and `years_observed` can be the literal string
`"None-None"` — 13 assertions, all `iia`, 70 rows. So the tool raised `ValueError` at import and the
triage queue **could not be regenerated at all**. That is the mechanism behind the staleness: the queue is
not stale because a step was skipped, it is stale because the step could not run.

### The fix keeps the affected assertions

`span()` returns `(None, None)`, and each of the three callers degrades to "no claim" at its own site — the
nesting screen cannot year-overlap them, a ranged alias cannot cover a segment that does not exist, and an
assertion with no span cannot sit on its candidate's boundary year. They stay **in** the queue behind a new
`span_missing` column. Dropping them was the tempting repair and would have made the crash's damage
permanent and invisible.

### A second defect, found because two counts disagreed

`NO_SPAN` counts distinct keys (12), the flag counts rows (13). A key is `label|source|years_observed`, so
a `None-None` span **stops disambiguating**:

```
ethiopia|iia|None-None  ->  ETH-1907-1936   pending, 1 row
ethiopia|iia|None-None  ->  ETH-1936-1941   pending, 1 row
```

Two assertions, two *different* candidate polities, one key. A key maps to one ledger row and one evidence
bundle, so banking a verdict writes to both (#308). `00_intake.py` **does** disambiguate colliding keys by
appending the candidate code and raises `SystemExit` if any survive — so a duplicate reaching this tool
proves the assertion set on disk predates that block. That is independent evidence for this issue's own
corrected reading: `assertions.json` is the stale side, not the queue.

### The gate had the same dead-guard shape, in code I wrote

`validate_assertion_triage.py` arms D and G only run when the span matches `(-?\d+)-(-?\d+)`. Those 13 rows
would have received **no year checks at all** while the gate reported success. The count is now printed and
pinned.

**The pin is 0, not 13.** My first version used 13, because that is what the current `assertions.json`
holds — which would have let the number regress to 13 while looking principled. Measured: a queue
regenerated by the documented production command has **0**, because `--period-col` takes it from 13 to 0
and `00_intake` now refuses to run without that flag on an input carrying `period` (#437). Zero is the
value a correct regeneration produces, so zero is the pin.

### The resync is not in this PR, and why that is the finding

Regenerating `assertions.json` with `--period-col` is a clear improvement:

```
null spans            13 -> 0
duplicate keys         1 -> 0
banked verdicts      155 -> 369     (+214: the keys finally match the ledger)
backlog           94,925 -> 33,882 rows
```

But the regenerated queue **fails arm G on four rows**:

```
india|fao1952|1938-1938   outside IND-1914-1937 (1914-1937)
brazil|iia|1913-1913      outside BRA-1903-1909 (1903-1909)
peru|iia|1913-1913        outside PER-1884-1909 (1884-1909)
cameroon|iia|1913-1913    outside GKM-1884-1912 (1884-1912)
```

Three are a **1909–1913 period collapsed to its end year** and routed to polities that cover the period's
*start*, because the row is dated by the period's END while `01` picks the polity by maximum COVERAGE of the
period. Which year represents a period average is #310's decision, not a drive-by change — and spans are
keys, so changing it moves banking.

So the crash fix is **necessary and not sufficient**: the resync is blocked on #310, not on the crash. That
is a more useful state than the issue had, where the resync looked like a step someone had declined to take.

All 71 gates pass; full 91-case selftest green. No tracked state is changed by this PR.


lbm364dl:
*Comment by Claude, correcting this PR's own description.*

The body says the four arm-G rows are blocked on "#310's decision about which year represents a period
average". **That is imprecise and I have measured why.** The periods are 1900–1913 and 1934–1938, and
`01`'s coverage rule picks the right polity for both (`PER-1884-1909` is live for ten of those fourteen
years; `IND-1914-1937` for four of five). What is wrong is the **span**: `00_intake` derives
`years_observed` from `eff_year`, which returns the period's END year, so the assertion says `1913-1913`
about a row placed using 1900–1913.

So it is not a period-semantics decision — it is two rules in one pipeline disagreeing, where the measured
rule is not the one that produced the key. Full working on #434.

Nothing in this PR changes as a result: the crash fix, the `span_missing` column, the collision report and
the zero pin all stand, and the resync is still correctly out of scope. What changes is the *reason* it is
out of scope — correcting the span re-keys **291 of 1,074 assertions** (28,069 rows) and orphans their
banked verdicts, which is a migration with a known absorber (`23_verdict_carryover.py`), not a blocked
decision.

Issue #431, opened 2026-08-19
The atomic-write lesson never propagated: 4 exposed write sites (not 18 tools), fixed in #439 and now gated by an AST sweep that finds 48 sites where a grep found 16


lbm364dl:
> ### ⚠️ Corrected — see [this comment](https://github.com/eduaguilera/whep-polities/issues/431#issuecomment-5341214604)
>
> **The "18 tools" figure below overstates the exposure by roughly four-fold.** It came from matching `*.csv` filenames anywhere in a file, so it counted files a tool merely **reads**. Verified per write site, the adjudication risk is **four write sites in two files** — `review_ledger.csv` (truncated by three separate tools) and `quarantine.csv` (one).
>
> Also wrong below: `verdicts_applied.jsonl` is opened in **append** mode and is not exposed; `select_tranche.py` reads it rather than writing it; `apply_verdicts.py` truncates **one** tracked file, not six; and no truncating write exists for `source_conventions.csv`, `applied_aliases.csv`, `suspect_wiki_pages.csv` or `wiki_notes_queue.csv`.
>
> The original text is kept below rather than edited away, because the method error — resolving a write target back to its path constant is the check that should have been there — is the more useful part.

*Filed by Claude, found by sweeping for repo lessons that were applied in one place and never propagated — the same search that found the blank-source alias wildcard.*

`16_source_splices.py` onward build their output fully in memory and then `os.replace` it, each carrying a comment to the effect that *a truncating `open(path, "w")` has cost this repo two tracked state files mid-error*. That lesson is applied in **every tool numbered 16 and above** and in **none numbered 14 or below**. It was applied forward and never back-propagated.

## Measured

Tools with no `os.replace` that write at least one **tracked** file under `pipelines/polity-autoimprove/state/`:

```
00_intake.py                     applied_aliases, review_ledger, source_conventions
01_match_and_findings.py         applied_aliases, review_ledger
02_territorial_evidence.py       iia_territorial_notes, review_ledger
04_territory_basis.py            footnote_flags, territory_basis
05_magnitude_screen.py           magnitude_impossible_area, magnitude_outliers
06_landuse_consistency.py        landuse_corrections
07_yield_consistency.py          impossible_pairs, yield_corrections, yield_series_corrections
10_livestock_consistency.py      livestock_corrections
12_subnational_sums.py           subnational_sums
12_trade_mirror_gap.py           trade_mirror_gaps, trade_mirror_summary
[...]

PR #461, opened 2026-08-19
Gate the atomic-write rule instead of trusting a grep: 48 write sites, not 16 (#431)

lbm364dl:
Addresses #431.

#431 found four truncating writes to unregenerable state files and #439 fixed them. **The detector that
found them was `grep open(x, "w")`, and the issue said in its own caveat that this is a floor, not a
census** — it cannot see `pandas.to_csv`, `pathlib.write_text`, or a `json.dump` whose `open` sits on
another line.

Re-swept with an AST: the repository has **48 truncating write sites, not 16**. The regex was blind to about
forty, nearly all `to_csv`. Every one of those forty writes a **regenerable** output, so nothing was actually
exposed — but that was luck, not design. A `to_csv(LEDGER)` added tomorrow would have been invisible to the
very check that found the original four.

### What is protected, and what deliberately is not

Ten files, each with its reason recorded in the registry: `review_ledger.csv` (three writers, two of which
read it, edit it in memory and write it back over itself), `quarantine.csv`, `applied_aliases.csv`,
`source_conventions.csv`, `data_errors.csv` and the rest. Regenerable measurement tables are **not**
protected, on the same ground `atomic.py` gives for skipping them: a truncation there is repaired by
re-running the tool, and a gate that overreaches gets weakened rather than obeyed.

**Append is not truncation.** `verdicts_applied.jsonl` is `"a"`-mode, and #431's first pass wrongly listed it
as exposed; the gate asserts it is never opened `"w"`.

### The rule is phrased negatively, so atomic writers need no special case

An atomic writer never names its destination in the truncating call — it truncates a temp file in the same
directory and `os.replace`s it. So the check looks for a truncating call whose target **resolves** to a
protected path, which an atomic writer cannot produce. `write_csv_atomic(LEDGER, …)` and a hand-rolled
mkstemp/replace pair both pass with no exemption.

### The selftest case is the one the old detector would have missed

It restores the ledger write as a **`to_csv`** — the call form the original grep could not see. If this gate
is ever reduced to a regex over `open`, that case goes green while `review_ledger.csv` is truncated on every
run of the tool that reads it. Verified: exit 1 naming `truncates review_ledger.csv`.

A coverage floor (`MIN_WRITE_SITES`) is included so a rewrite that stops parsing, and therefore stops
checking, cannot pass for the wrong reason.

Two files a first pass flagged — `label_alias_map.csv` and `source_flow_flags.csv` — are `data/final/`
derived maps, regenerable by re-running their writer, so neither belongs in the protected set.

Gates 73 → 74, selftest 94 → 95. All 74 gates pass locally; full 95-case selftest green.


lbm364dl:
*Comment by Claude.* Corrected two numbers in the description above: I wrote **"Gates 74 → 75, selftest
96 → 97"**. Verified per branch:

```
main                                validate_*.py 73   selftest CASES 94
this branch                         validate_*.py 74   selftest CASES 95
```

So it is **73 → 74** and **94 → 95**. I had used the post-change count as the "before" — 74 is what this
branch has *after* adding the gate, and 96 is the count on a separate follow-up branch that adds a case to
`validate_collapse_groups.py`. Neither figure changes anything about the gate itself; both were simply
unchecked when I wrote them.

Issue #416, opened 2026-08-19
IIA tobacco and hops are ~100x too large from 1934: 427 rows across 93 labels, 348 convicted per-row, and existing corrections reach 28


lbm364dl:
*Filed by Claude. #377 established the x100 factor and #414 gave the mechanism a home; this establishes the **scope**, which is 13x larger than what is currently corrected.*

## The break is at 1933/1934, and it is exactly where the second opinion runs out

Layer-B `iia` tobacco/hops, production ÷ area, per year. Tobacco yields ~0.5–3 t/ha:

```
1909-1933   319 pairs   median  0.90 t/ha    1% above 20 t/ha
1934-1945   191 pairs   median 58.90 t/ha   84% above 20 t/ha
```

Year by year the transition is a step, not a drift: 1930 `0.80`, 1931 `0.72`, 1932 `0.70`, 1933 `0.77`, then **1934 `50.38`, 1935 `50.17`, 1936 `62.27`, 1937 `67.80`**, staying there through 1945.

**1933 is the last clean year, and #414 says why.** It is the only year covered by two yearbook volumes. Where `iia_1933_34` offers a second opinion the pipeline takes it; from 1934 on, `iia_1938_39` and `iia_1939_45` are the sole source and the inflated values pass straight through. Identical mechanism to #414's false zeros, same boundary, same cause.

## Confirmed independently by the volume comparison

From `state/edition_conflicts.csv` (#415), the 1933 overlap:

```
revisions >= 50x between volumes                89    ALL in 1933
  larger value held by iia_1938_39              88 of 89
  of which tobacco (63) or hops (16)            79
tobacco/hops revisions with ratio 80x-130x      77    larger value in iia_1938_39: 77 of 77
```

| cell | iia_1933_34 | iia_1938_39 |
|---|---|---|
| turkey / tobacco / production 1933 | 35,367 t | **4,014,800 t** |
| dutch east indies / tobacco 1933 | 27,829 t | **3,255,800 t** |
| british nyasaland / tobacco 1933 | 1,582 t | **286,200 t** |
| total general / hops 1933 | 47,600 t | **4,920,000 t** |

Turkey produced roughly 35–60 kt of tobacco in the 1930s. The two methods — physical yield and inter-volume comparison — agree on the boundary and on the direction, and neither depends on the other.

## The scope, and why the current detector cannot reach it

```
iia tobacco/hops production rows (tonnes)        696
  1934 onward                                    328   across 56 labels
[...]

PR #465, opened 2026-08-19
The tobacco era is 427 rows across 93 labels, not 328 across 56: period rows were dropped silently (#416)

lbm364dl:
`29_era_shift_verdicts.py` filtered `prod.year >= ERA_FROM`. **That is False for NaN**, so every
period-average row left no trace. 341 of this source's 1,037 tobacco/hops production rows carry a period
instead of a year, and **99 of those are `1934-1938`** — squarely inside the era.

```
scope        328 rows / 56 labels   ->   427 rows / 93 labels
convicted    266 (81%)              ->   348 (81%)
   impossible_yield            160  ->   232
   no_area_level_shift          56  ->    66
   impossible_yield_zero_area   50        unchanged: all dated
```

**The convicted share holding at exactly 81% is the reassuring part.** The period rows behave like the dated
ones, so they are the same defect rather than a different population that happened to be excluded.

### The same blindness as #464, found the same day

#464 corrected #456's "nothing else moves" because my diff measured on the raw `year` column while the
pipeline reasons over a period's **span**. This is the identical mistake in a screen I wrote a few hours
earlier. Both are now stated at the site rather than merely fixed, because the shape recurs: **a comparison
against a year silently excludes every row whose year is null, and null-year rows are 5.12% of the panel.**

### How period rows are handled

They pair with an area average of the **same span**, keep an **empty `year`**, and carry their `period` in a
new column — so a five-year mean can never be read as an observation of one year. Arm E now requires exactly
one of `year` or `period` and checks that whichever is present reaches the era. A period is included when it
**ends** in the era: `1934-1938` is entirely inside, `1928-1932` entirely before, and nothing in this source
straddles the boundary.

Verified: giving a period row a year as well fails arm E naming both fields; the zero-area selftest case
still bites; 74 gates and the full 96-case selftest pass.

### Bound restated

These rows are excluded from publication anyway (#310, documented in #460), so this is evidence about the
era's **extent** — 37 more labels affected — not about published figures. The `impossible_yield_zero_area`
count is unchanged at 50 because all of those are dated.

### DIFF
--- a/pipelines/polity-autoimprove/29_era_shift_verdicts.py
+++ b/pipelines/polity-autoimprove/29_era_shift_verdicts.py
@@ -56,6 +56,7 @@ from __future__ import annotations
 import argparse
 import csv
 import os
+import re
 import sys
 import tempfile
 
@@ -72,7 +73,7 @@ IMPOSSIBLE_YIELD = 20.0         # t/ha; an order of magnitude above the crop's c
 HIGH_YIELD = 3.0
 LEVEL_SHIFT = 30.0              # x the label's own pre-era median
 
-FIELDS = ("source", "label", "whep_code", "item", "year", "unit", "production", "area_ha",
+FIELDS = ("source", "label", "whep_code", "item", "year", "period", "unit", "production", "area_ha",
           "implied_yield", "own_pre_era_median", "ratio_to_own", "verdict", "convicted")
 
 CONVICTING = {"impossible_yield_zero_area", "impossible_yield", "no_area_level_shift"}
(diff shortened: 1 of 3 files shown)


lbm364dl:
*Comment by Claude.* I swept the whole pipeline for the shape this PR fixes — a filter written as a `year`
comparison, which is silently False for the 5.12% of panel rows whose year is null — so the class is bounded
rather than left open. **Four sites, and the two I have not touched are measured immaterial.**

```
29_era_shift_verdicts.py:125   prod.year >= ERA_FROM        THIS PR: 328 -> 427 rows in scope
29_era_shift_verdicts.py:103   prod.year <= ERA_FROM - 1    follow-up: 50 labels had no baseline
11_retest_conventions.py:122   year >= 1922 & <= 1940       immaterial, measured below
11_retest_conventions.py:165   year >= 1949 & <= 1960       immaterial: zero period rows exist
```

**`check_iia_russia`** would take 104 more rows if period averages were included, and its medians barely
move:

```
cotton lint  ha    dated-only 2,033,000   with periods 1,990,500     threshold > 1,000,000
rye          ha    dated-only 24,026,000  with periods 24,656,448    threshold > 15,000,000
```

Both clear their thresholds by a wide margin either way, so the check's conclusion is unaffected and I have
not changed it. One incidental note for whoever revisits that convention: the figures quoted in
`source_conventions.csv` (2,033,000 and 24,026,000) are the **dated-only** medians, so a recomputation that
included periods would differ slightly without anything being wrong.

**`check_iia_juan_germany`** has **zero** period rows in its window, so the filter cannot exclude anything.

That leaves the two sites in this file, both mine, both from today — plus #464, where the same mistake was in
my *verification* of #456 rather than in a screen. Three real instances, one bounded negative result.

Issue #22, opened 2026-07-27
Long single-vintage rows overstate early-period territory (Cape, Burma, Paraguay, Indonesia)


lbm364dl:
Several polity rows span decades with a single polygon vintage, so data from the early years is matched to a territory that did not yet exist. The approximations are documented, but each is a real, quantified error worth deciding on.

**Known cases, with the source situation:**

| polity | span | issue | source available? |
|---|---|---|---|
| `CAP-1800-1910` | 110y | Cape Colony annexed British Kaffraria, Griqualand West (1880), Transkei (1885-94), British Bechuanaland (1895). Polygon **overstates the early decades**. Assertion banked `best_available`. | CShapes coverage starts 1886 — earlier boundaries need another source |
| `IND-1800-1886` | 86y | the entire British conquest (see its own issue) | none |
| `IDN-1800-1945` | 145y | Dutch control still consolidating: Aceh War from 1873, outer-island campaigns into the 1900s | CShapes constant 1886-1945 |
| `MMR-1885-2025` | 140y | one row across British Burma, independence and the 1989 renaming; verification banked it `best_available` | not investigated |
| `PRY-1870-1932` | 62y | geometry includes Chaco Boreal, only awarded in 1938 — **overstates by ~87%** (mostly sparsely populated) | `estimate`, documented |

`04_territory_basis.py` already classifies every polity as `measured / assumed_constant / back_projected / unassigned` and writes `state/territory_basis.csv` — that is the tool for finding the rest systematically rather than case by case.

**The general question worth settling:** when a row's territory demonstrably changed mid-span but no source can split it, is the right answer (a) split anyway with a documented estimate for the earlier period, (b) keep one row and quantify the error on the page, or (c) mark the row's early years as unusable for area-normalised analysis? Today's practice is (b), which is honest but leaves every downstream user to rediscover the caveat.

Related: the `constant_territory` back-casting work in the WHEP R package deals with the same problem from the analysis side, so a shared convention would help both.

lbm364dl:
> *Filed by Claude (Claude Code). Measured.*

**A case for this issue, found while clearing a flag on a different one (#208, #123) — and it is this issue's premise inverted.**

`MOR-1800-1904` spans 104 years on one Cliopatria vintage. Five vintages fall inside it:

```
1769-1860   387,614 km²   <- bound
1861-1869   386,677
1870-1884   386,205
1885-1889   358,263
1890-1904   358,422
```

[...]

PR #499, opened 2026-08-20
Retarget 17 lexicon exonyms to polity names: stated-area coverage 205 → 223 polities, and Newfoundland turns out to include Labrador (#195)

lbm364dl:
*PR by Claude.* All 80 gates green, 103 selftest cases pass, `--check` regenerates clean.

#195's problem is that check A compares a declared area against the geometry it was copied from. The
non-circular alternative — comparing against what the **source states** — is already wired; the only
limit is reach: **205 of 716** live polities, because **1,190 of 2,225** stated-area statements never
resolve.

## The cause was not missing translations

**17 of the failing labels already had lexicon entries**, correct as history and wrong as lookups: they
map each French label to its English **exonym**, while the target vocabulary is the database's
`polity_name`.

```
ceylan              'Ceylon'          ->  Sri Lanka
cote de l or        'Gold Coast'      ->  Ghana (1898-1957)
terre neuve         'Newfoundland'    ->  Dominion of Newfoundland
guyane britannique  'British Guiana'  ->  Guyana
zanzibar            'Zanzibar'        ->  Zanzibar Protectorate
alaska              'Alaska'          ->  Territory of Alaska
formose             'Taiwan'          ->  Formosa
```

`formose` is the one that resolved to something **wrong** rather than to nothing: *Taiwan* matches
`TWN-1945-2025` while its statements are 1911–1929, which belong to `TWN-1895-1945 "Formosa (Japanese
Taiwan)"`. Two retargets are historically load-bearing — `finlande` → *Grand Duchy of Finland* and
`hongrie` → *Kingdom of Hungary (Transleithania)*, which is what those 1911–1913 statements describe.

## Every retarget was verified against its own years before being written

A candidate was kept only if it resolves for **every** year that label carries, so no entry trades one
covered year for another. 28 labels passed, **0 partial, 0 failing**.

```
basis table   286 statements / 205 polities  ->  310 / 223     (21 polities newly covered)

HUN-1800-1918   325,411 stated  vs  325,419 polygon   1.000
DEU-1871-1919   540,857             540,041           0.998
TWN-1895-1945    35,846              35,994           1.004
LKA-1800-2025    65,608              65,955           1.005
```

A mis-targeted label would land nowhere near 1.0, so those ratios are independent confirmation of the
retargets rather than a coincidence.

## Two real divergences surfaced — baselined with their reason, not suppressed

**`NFL-1907-1949` is the substantive one.** IIA states **110,679 km²**, the *island* of Newfoundland; our
polygon is **398,115 km²**, the island **plus Labrador**. Both are correct for what they describe — the
Dominion did administer Labrador — so the polygon is unchanged. But a per-km² intensity computed from a
yearbook numerator over this polygon **understates by 3.6×**, with nothing else in the repo to warn of it.
That is the Algeria case again, and the largest of its kind after it.

`MAC-1800-2025`: IIA states 14 km², our polygon is 34 km². Macao roughly doubled by land reclamation, so
this is a **vintage** difference rather than a scope one — the #22 shape at a magnitude this source cannot
resolve any other way.

## One stale baseline removed, and one ceiling raised for a reason worth reading

`("SMR-1800-2025","fao")` recorded a real divergence for a right reason — FAO's 1,000-hectare grid can
only record 61 km² as 100 — but IIA's nine statements now supply **61 against our 61** (1.003×), so the
polity-level figure agrees and the entry baselines something that no longer exists. The FAO row itself is
untouched and still reads 0.612× with its note.

`BASELINE_AVOIDABLE_SELF_REF` **28 → 33**. A raised ceiling normally means drift; here the causation runs
the other way and the comment says so. Nothing declared a new self-referential area — the ceiling counts
rows that *could* have compared against a source, so it grows when the **source** side grows. 5 of the 21
newly-covered rows already declared their own geometry's area, so the arm **gained reach over 5 rows**
rather than losing ground on 5, and it now reports on 33 rows where check A is silent while a yearbook
figure exists.


lbm364dl:
*Comment by Claude.* **Correcting this PR: three of the 17 retargets were a swap, not a gain, and are now
reverted.** The body above claims *"no entry trades one covered year for another"* — that was false for
`allemagne`, `finlande` and `hongrie`.

I found it by re-running the unresolved census after the change, where all three reappeared for
**1921–1937**. The cause is a one-directional test: I verified each candidate against the years its label
was **failing** on, and never against the years it was already **succeeding** on.

```
lost DEU-1920-1938 (6 statements)    gained DEU-1871-1919 (3)
lost FIN-1917-1940 (6)               gained FIN-1809-1917 (3)
lost HUN-1920-1938 (6)               gained HUN-1800-1918 (3)
```

Identical polity count either way and 9 fewer statements resolved, so the historically-nicer target is
strictly worse. All three are reverted to `Germany` / `Finland` / `Hungary`, and each entry now carries a
note saying why it is *not* retargeted: the better target is right for 1911–1913 and wrong for 1921–1937,
and a flat `label -> name` lexicon cannot express a target that depends on the data year. That is the
schema limit I flagged on the issue, and these three are its concrete instances.

**Revised result:** 205 → 223 polities, **zero lost** (was 21 gained / 3 lost), 286 → 310 (polity, source)
bases. The 14 surviving retargets are unaffected, and `NFL-1907-1949` (island vs island-plus-Labrador,
3.6×) and `MAC-1800-2025` are unchanged — those remain the substantive findings.

`BASELINE_AVOIDABLE_SELF_REF` is **32**, not 33: the revert moved the true count, and a ceiling one above
the measurement would let a real regression through unnoticed. The comment records that it was briefly 33
and why.

PR #508, opened 2026-08-20
Reunion's tobacco baseline is itself inflated, so 10 era exonerations are false (#416)

lbm364dl:
*PR by Claude.* Documentation only; all 81 gates green.

`era_shift_verdicts.csv` convicts a no-area row when it exceeds **30×** the label's own pre-1934 median.
For `reunion` that median is **20,000 t and is itself inflated**, so every era row divides one inflated
number by another and reads as consistent — `ratio_to_own` 1.00 at 1934/1935/1936 and 0.76–1.11 for
1939–1945. All ten filed `no_area_level_consistent`, none convicted.

```
raw french reunion / tobacco / production
  1926     423.1 t   iia_1929_30
  1927     348.4 t   iia_1929_30
  1933  20,000.0 t   iia_1938_39     <- becomes the pre-1934 baseline
  1934-36  20,000    iia_1938_39
  1939-45  15,200 - 22,300           iia_1939_45
```

A few hundred tonnes to 20,000 is a **47× step at 1933**, and Réunion is 2,512 km² — the hundreds are
plausible and the 20,000 is not.

## Why the 1933 second opinion did not rescue it

`iia_1933_34` carries **no dated 1933 production** for this label — only an undated 670.8 t row — so there
was nothing to outvote the late volume. The inflated figure filled 1933 and became the baseline for every
later year.

The general condition, which is the transferable part: **a 1932–33 baseline is clean only where
`iia_1933_34` supplies a dated value for that label and item.** Elsewhere the level-shift test compares
the defect against itself.

## Recorded rather than repaired, deliberately

The verdict table is left alone. Its verdicts are derived and its `--check` is current, so a criterion
change means regenerating a tracked artefact of adjudications — the operation that has already gone wrong
here once, re-banking confirms over hand retractions from a stale file.

## Not a threshold problem

The other ten `no_area_level_consistent` rows sit at **3.2× to 27.1×** their own baseline and are simply
under the 30× line (`jordan` 27.1, `mauritania` 25.0). That is a question about where a conservative line
sits, not about a contaminated comparison — and it wants a different answer. The 348 convictions are
unaffected either way.


lbm364dl:
*Comment by Claude.* **Correcting this PR's stated condition, and adding an eleventh case.** I diagnosed the
ten Réunion rows and left `romania / hops` — the other near-ratio row — undiagnosed. Checking it refutes
the condition I wrote.

Romania is contaminated the same way, **but `iia_1933_34` does supply a clean dated 1933 for it**:

```
raw romania / hops              iia_1933_34            iia_1938_39
  1930   29.9 t on 71 ha        clean
  1931   48.3 t on 85 ha        clean
  1932   15.1 t on 29 ha        clean
  1933    9.5 t on  2 ha        clean            900 t on 1,000 ha   inflated
  1934-1937                                      800 / 2,400 / 1,000 / 4,200 t
  period 1925-1929   40.6 t     clean
  period 1928-1932            3,500 t   <- iia_1938_39, inflated
```

So "no dated 1933" is **not** the condition. The arithmetic says what is: the recorded baseline of
**1,770** is exactly `(3,500 + 40.6) / 2` — the mean of the clean `1925-1929` period row and the inflated
`1928-1932` one. **The era baseline includes pre-era PERIOD rows**, so one inflated period row contaminates
it however many clean dated years sit beside it.

Corrected condition: a pre-1934 baseline is clean only where **every** pre-era row feeding it — dated or
period — comes from a volume without the ×100. Réunion needed a missing 1933; Romania did not, and is
contaminated anyway.

The entry is widened to both labels and eleven rows. The other ten `no_area_level_consistent` rows remain a
threshold question (3.2×–27.1× under a 30× line), and the 348 convictions remain unaffected.

Issue #433, opened 2026-08-19
Map: six IIA defect classes found on 2026-08-19 share two volumes and one era, and three of their remedies interact


lbm364dl:
*Filed by Claude. Not a new defect — a map. Nine issues were filed or substantially extended today and they are not independent: they concentrate in two yearbook volumes, they meet in the same cells, and at least three of the remedies interfere with each other. Anyone acting on one should see the others first.*

## The volumes, measured

```
production+area zero rate, by volume
  iia_1938_39   6.45%      <- 14x to 60x every other volume
  iia_1909_21   0.82%
  iia_1939_45   0.39%
  iia_1925_26   0.12%
  iia_1933_34   0.11%
  iia_1929_30   0.10%

zero contradicted by another volume : 83 of 83 hold the zero in iia_1938_39
power-of-ten disagreements          : 98 of 99 hold the SMALLER value in iia_1933_34
impossible pairs (area=0, prod>0)   : 176 of 177 fall in 1933-1938, iia_1938_39's window
```

`iia_1933_34` is the smaller side 55% of the time across *all* revisions between it and `iia_1938_39` — a coin flip — so the 98-of-99 split is a property of the power-of-ten class, not of the volume being generally lower.

## The six classes

| # | class | volume implicated | scale |
|---|---|---|---|
| **#414** | blank cells read as 0 | `iia_1938_39` (60%), all five contribute | 770 raw, ~539 reaching layer B |
| **#416** | tobacco/hops ×100 from 1934 | `iia_1938_39`, `iia_1939_45` | 328 rows, 252 per-row convictable |
| **#424** | ×10 dropped digit at 1933 | `iia_1933_34` holds the low value | 33 cases, 14 confirmed in layer B |
| **#426 / #427** | meslin↔spelt and calcium cyanamide↔nitrate transposed | `iia_1938_39` | 8 + 4 labels, no propagation measured |
| **#420** | area = 0 beside positive production | 1933–1938 | 177 cells |
| **#422** | Karafuto routed to Russia | not volume-specific | 21 enumerated cells |

## Why 1933 keeps appearing

[...]

Issue #446, opened 2026-08-19
The panel has no channel for value precision: 89,938 non-zero values (47.6%) sit on a 1000-grid and read as exact


lbm364dl:
*Filed by Claude.*

Two separate findings this week ended at the same missing thing: **the panel has no channel for value
precision**, so a figure rounded to the nearest 100 is indistinguishable from an exact one.

## The exposure, measured

PR #445 recorded the reporting grid of the IIA yearbook volumes as a convention. The volumes cover
disjoint year windows (except the 1932-33 overlap that gives the pipeline its second opinion):

```
iia_1909_21  1909-1921      iia_1933_34  1929-1933
iia_1925_26  1920-1925      iia_1938_39  1932-1938   <- 100-grid: 95.0% of non-zero values
iia_1929_30  1926-1929      iia_1939_45  1939-1945   <- 100-grid: 95.6%
```

and the coarseness propagates all the way into the published panel:

```
layer B iia rows                                        26,175
  from 1934 onward (the coarse-volume window)           11,377   (43% of iia, 5.9% of all layer B)
  of those, non-zero and sitting on a 100-grid           9,962   (92%)
```

**So ~9,962 published values carry roughly ±50 units and are presented as points.** A `1,000` from 1940
and a `1,000.4` from 1925 look identical to every consumer, and only the second is a measurement of
anything finer than a hundred.

## The two findings that need this

- **#366 (constant runs).** 1,619 rows repeat one value across years. I filed it arguing they could not be
  a resolution limit; that argument is withdrawn — the finer evidence is cross-era in 169 of 170 cases, and
  the run constants (100; 1,000; 2,000; 3,000; 5,000) are exactly what survives this grid. The rows are not
  wrong, but a consumer reading them as "this did not change" is misled. **No value repair applies; the
  only honest fix is to publish the precision.**
- **#433 (impossible-pair repair).** The `production / reference_yield` repair for the 177 zero-area cells
  divides by a rounded numerator in **167 of 177 (94%)** cases. Because those productions are small (median
  300 t), the grid imposes a median **16.7%** relative uncertainty on the implied area, over 1% in 143 of
  177. The repair is still better than a zero — but it produces an interval, and there is nowhere to put
  the interval, so it would be published as a point.

Both remedies are the same remedy, which is the argument for treating it once rather than per-issue.

## What a fix would look like

Not a patch, because this is a layer-B schema question and `consolidate_layer_b.py` lives outside this
[...]

PR #510, opened 2026-08-20
Hops carries a SECOND defect — its area is x10 — which reconciles the 70x/21x yield figures (#416)

lbm364dl:
*PR by Claude.* Documentation only; all 81 gates green.

The `iia-hops-x100` entry treats hops as a production-only ×100. Measured on same-year 1933 pairs, where
`iia_1933_34` and `iia_1938_39` both report:

```
hops     production   n=15   median 100.00x   15/15 within 20% of 100x
hops     area         n=16   median  10.04x   14/16 within 20% of 10x
tobacco  area         n=46   median   1.00x   32/46 within 5% of 1x     <- clean
```

Across **every** item, **zero** area pairs fall in the 30–300× band. So the ×100 is production-only, and
the hops area ×10 is a second, distinct defect.

## The late volume also rounds it, which explains the only two outliers

```
hungary   107.0 -> 1,000     usa      12,262.0 -> 123,000
france  1,708.0 -> 17,000    austria      56.0 ->   1,000    17.86x
uk      6,837.0 -> 68,000    romania       2.0 ->   1,000   500.00x
```

Every late value is a multiple of 1,000 against precise clean values, so the unit is ten times smaller
**and** rounded to the nearest 1,000 of it — the nearest 100 ha. `austria` and `romania` sit *below* that
floor, so their 17.9× and 500× are the floor, not a third defect. Same phenomenon
`validate_grid_ambiguous_zeros.py` gates.

## It reconciles a figure that has been on record without an explanation

#433 records tobacco breaking at **70×** and hops at **21×** on implied yield. Yield is production over
area, so tobacco inflates 100/1 and hops **100/10 = 10**. The items differ on yield *because* hops' area
is inflated too — which also means a yield-based screen understates hops tenfold, likely why
`yield_series_corrections.csv` reaches so few hops series.

## Exposure and remedy

41 layer-B hops area rows at 1934+, median 4,000 ha against a pre-1934 median of 271, and **22 of the 41
are exact multiples of 1,000 against 1 of 80** before 1934 — a signature no genuine area series produces.

Remedy: hops production ÷100 **and** hops area ÷10, with values under ~100 ha unrecoverable because the
source rounded them to its floor. Dividing hops production alone would leave the area ten times high and
the yield ten times low — worse than leaving both, because it would look repaired.


lbm364dl:
*Comment by Claude, correcting part of this PR's evidence before it merges.* **One of the two exposure
arguments does not hold: the "multiples of 1,000" signature is a general property of the late volumes, not
something specific to hops.**

I wrote that 22 of 41 hops area rows at 1934+ are exact multiples of 1,000 against 1 of 80 before 1934,
and called that "a signature no genuine area series produces." Measured across every item:

```
ALL iia area rows 1934+ : n=3,393, on a 1,000-grid: 96%

share of area values on a 1,000-grid, <=1933 -> >=1934
  jute                      2% -> 100%      rye                    5% -> 100%
  tobacco                   4% -> 100%      cotton lint           10% -> 100%
  soybeans                  4% -> 100%      grapes                10% -> 100%
  (27 items measured, median jump 91 percentage points)
```

That is the **reporting grid** already documented for these volumes (#446, #455: `iia_1938_39` puts 95% of
non-zero values on a 100-grid), not a defect signature. Hops is in fact *below* the field at 50% on-grid
against 96% overall, so the grid says nothing in its favour.

## What still stands, and what is now an inference

**Stands:** the ×10 itself, measured same-year at 1933 where both volumes report — 15 of 16 hops area
pairs, median 10.04×, with the two outliers explained by the rounding floor. That is a direct measurement
and it is unaffected.

**Now an inference rather than a measurement:** that the ×10 continues into 1934+. It is a reasonable one
— the same volume, the same column, and no reason for it to change mid-run — but the panel's median
jumping from 271 to 4,000 ha is confounded by composition (different countries reporting in each period),
and the grid share carries no information as shown above. So the **41 affected rows** should be read as
"41 rows in the window where the defect was measured at its boundary", not as 41 rows independently shown
to be inflated.

## Also worth recording: the ×10 area defect is not confined to hops at 1933

Sweeping the 5–30× band for area, same year, both volumes:

```
hops        15 of 16 pairs   median 10.04x    <- near-universal
cacao: raw   3 of  8         median 10.00x
rye          3 of 39         median 10.00x
plus singletons across cottonseed, groundnut, rapeseed, sesame, soybean, spelt, tea  (16 cells total)
```

All 16 late values are multiples of 1,000 against precise clean values, so they fit the same
×10-then-rounded model. But for the two I could trace, **layer B took the clean value** (`cacao` jamaica
390, `soybean` japan 32,367), so the second opinion rescued them and they are a raw-level defect the
pipeline already handles at 1933. Hops is the one where near-universality at the boundary makes the 1934+
continuation worth acting on.

I will amend the entry's exposure sentence accordingly rather than leave the stronger claim on record.

Issue #424, opened 2026-08-19
A x10 class at 1933 sits below every detector's floor: 14 layer-B cells are ten times too small, and the source prints the right value in another volume


lbm364dl:
*Filed by Claude. Found by testing `state/edition_conflicts.csv` (#415) for factors that are clean powers of ten — the same test `07_yield_consistency.py` applies to yields.*

## The finding

Of 1,058 inter-volume revisions, **99 are within 5% of a clean power of ten**, and **98 of the 99 hold the smaller value in `iia_1933_34`**.

The control matters here and it is decisive: across **all** 1,032 revisions between `iia_1933_34` and `iia_1938_39`, the 1933-34 volume is the smaller side **55% of the time** — a coin flip. Among the power-of-ten cases it is smaller **98 of 99 times**. That is not a property of the volume being generally lower.

The exponents split cleanly by commodity:

| | cases | products |
|---|---|---|
| **×100** | 65 | tobacco (52), hops (13) — **only** these two |
| **×10** | 33 | 12 products: hops, sugar cane, cacao, cottonseed, rye, sesame, groundnut, meslin, spelt, eggs, tobacco, tea |
| ×1000 | 1 | rye |

The ×100 set is **entirely** #416's tobacco/hops inflation, arriving from the volume side and confirming it. **The ×10 set is new.**

## The ×10 cases are a dropped digit in `iia_1933_34`, and 19 of 20 are at 1933

```
denmark  / rye area 1933            14,302   vs   143,000
spain    / rye area 1933            59,073   vs   591,000
sweden   / rye area 1933            22,103   vs   226,000
japan    / soybean area 1933        32,367   vs   324,000
peru     / cottonseed area 1933     13,048   vs   130,000
uganda   / groundnut area 1933       5,103   vs    51,000
```

Note the shape: `iia_1933_34` carries a **precise** figure and `iia_1938_39` a **rounded** one ten times larger. Denmark's rye area really was around 140,000 ha, not 14,302; Spain's around 590,000.

## Why nothing catches them

`iia/japan/soybeans/ha` is the clearest case, because its own neighbours convict it:

```
1930: 346,750   1931: 350,347   1932: 341,753   1933: 32,367   1934: 336,000   1935: 333,000
```

[...]

Issue #490, opened 2026-08-20
Eight iia fertilizer series publish a MINOR material as the nutrient category: spain/p is phosphate rock at 5,400 t while the same page prints 999,607 t of superphosphate


lbm364dl:
*Issue created by Claude.*

#378 found that an `iia` item series can switch raw product year to year, and #379 gated it on
**oscillation** (A…B…A) because a one-way switch is indistinguishable from growth. That criterion is
sound, and it has a blind spot: a series that **never switches** because it consistently picks the
*smallest* of several materials the source prints. No oscillation, no level shift, no seam, no
magnitude outlier — every value is plausible for the product it came from. Eight series are in this
state, and only one of them (`japan / p`) appears in #379's gated table.

## The exhibit: spain / p / tonnes

Both products, same volume, same country, same year, same unit:

```
raw iia_1933_34, spain, production, tonnes
  1930   fertilizers: calcium superphosphate   999,607      fertilizers: phosphate, natural   5,400
  1931   fertilizers: calcium superphosphate   887,850      fertilizers: phosphate, natural   7,734
  1932   fertilizers: calcium superphosphate   994,185      fertilizers: phosphate, natural   9,980
  1933   fertilizers: calcium superphosphate   966,653      fertilizers: phosphate, natural  14,507

layer B  iia / spain / p / tonnes
  1930  5,400     1931  7,734     1932  9,980     1933  14,507
```

Layer B carries the rock and drops the superphosphate — **15 of 15 attributable cells**, worst 185×.
The choice is the ingest's, not the source's: the larger figure is on the same page.

## All eight

| series | under-selected / attributable | worst | picked, and what stood beside it |
|---|---|---|---|
| `japan / p` | 7/8 | **3138×** | 1930: phosphate rock 305 vs 957,159 |
| `latvia / p` | 4/4 | 381× | 1930: basic slag 123 vs 46,918 |
| `spain / p` | 15/15 | 185× | 1930: phosphate rock 5,400 vs 999,607 |
| `germany / n` | 13/14 | 60× | 1909: calcium cyanamide 5,500 vs 330,450 |
| `france / n` | 11/15 | 60× | 1913: calcium cyanamide 1,246 vs 74,500 |
| `czech republic / p` | 4/4 | 32× | 1931: bone superphosphate 3,457 vs 112,198 |
| `norway / n` | 10/13 | 25× | 1931: calcium nitrate 14,999 vs 380,574 |
[...]

Issue #443, opened 2026-08-19
`unattributable` hides a splice class: 8 iia series are two raw labels in sequence, and USA sugar publishes a different COMPONENT in each of four eras (never a national total)


lbm364dl:
*Filed by Claude.*

`20_item_provenance.py` classes 142 `iia` item series as `unattributable`. That status reads as "we
could not tell which raw label this came from", but for at least **7 of them the answer is that it came
from two**, in sequence, inside one published series. The tool picks a single best-matching raw label,
so a series whose scope changes mid-run has no best match and falls into `unattributable` by
construction — the status hides the finding rather than reporting it.

## Method

For each `unattributable` series, match its layer-B `(year, value)` cells against every raw label's
cells, take the top two disjoint matches, and keep only those whose matched years are **temporally
separated** (no interleaving). Require ≥6 distinct matched values on **both** sides.

22 series show a clean temporal split; 7 survive the two-sided distinctness floor:

```
layer_b label        item                 unit    early raw label / years        late raw label / years
united states        sugar raw centrifugal t       usa                 1909-1921  usa louisiana and florida 1922-1945
russian federation   sugar raw centrifugal t       russia in europe    1909-1916  ussr                      1930-1940
russian federation   tobacco unmanufactured t      russia in europe    1909-1916  ussr                      1923-1936
malaysia             coffee green         ha      british federated malay 1909-1920  british malaya         1930-1944
serbia               yarn of true hemp    t       kingdom of serbs, croats 1909-1925 yugoslavia             1930-1938
syrian arab republic eggs hen in shell    t       french syria and lebanon 1931-1938 french syria           1939-1944
syrian arab republic grapes               ha      french syria and lebanon 1930-1938 french syria           1939-1945
```

I discarded 15 of the 22 on the floor, including `jamaica cotton lint -> bulgaria` (3 distinct values on
the Bulgaria side) which is plainly coincidence. **A note on one filter I got wrong:** I first also
rejected any pair whose matched values were mostly round numbers, which cut the list to 2. That filter
is invalid here — these yearbooks report to the nearest 100 tonnes, so `round=100%` is source precision,
not coincidence. The discriminator that works is distinct-value count, because a match requires the year
*and* the value to coincide.

## The USA series is the one to look at first

It does not merely switch once. It **alternates between two scopes**, and the layer-B series carries
both:

```
1909-1921   528,298 ... 1,043,651     national scale        [usa]
[...]

Issue #414, opened 2026-08-19
IIA reads blank cells as 0 across all volumes, worst by far in iia_1938_39: 770 false zeros there, ~539 reaching layer B, 83 contradicted by another volume


lbm364dl:
> ### ⚠️ Corrected — see [this comment](https://github.com/eduaguilera/whep-polities/issues/414#issuecomment-5340588510)
>
> **Two figures below are superseded.** The "87 of 88" is a loose per-**row** count of mine; the committed table and its gate report **83**, which is the defensible number.
>
> 88 covered zero rows collapse to 85 distinct **cells** (three cells carry duplicate zero rows), of which 84 are contradicted — and one of those is not a clean false zero at all: `british nyasaland / cottonseed / area / 1933` reads **0 *and* 12,000 in the same volume**, so the tool correctly classes it an ordinary revision rather than a contradicted zero. The tool is right and I was not.
>
> Also superseded: the title's original claim that this is one volume's behaviour. Traced per volume, `iia_1938_39` holds ~60% and **all five volumes contribute** — the rate anomaly is that volume's, the practice runs through the whole publication.
>
> The 770 zeros, the ~539 reaching layer B, and the 6.30% vs 0.04–0.82% rate gap all stand.

*Filed by Claude, found while testing #366's constant runs against the raw IIA extract.*

The raw extract carries a `yearbook` column naming the volume each row came from. Six volumes. Their zero rates on production and area rows are not comparable:

```
iia_1909_21    8,635 rows   zeros  71 = 0.82%
iia_1925_26    5,798 rows   zeros   7 = 0.12%
iia_1929_30    8,297 rows   zeros   8 = 0.10%
iia_1933_34   10,047 rows   zeros  11 = 0.11%
iia_1938_39   14,565 rows   zeros 940 = 6.45%     <-- 6x to 60x every other volume
iia_1939_45   16,611 rows   zeros  65 = 0.39%
```

**940 of the extract's 1,102 production/area zeros — 85% — come from one volume.** Nothing about 1933–1938 makes agricultural output sixty times more likely to be nil than in 1929–1932 or 1939–1945.

## Another volume contradicts them, 87 times out of 88

The decisive test, since some years are covered by two volumes: for each cell where `iia_1938_39` reads 0, does another volume carry a value for the same `(label, product, variable, unit, year)`?

```
zeros in iia_1938_39 (production + area)   770
  same cell also covered by another volume  88
[...]

PR #611, opened 2026-08-31
issues 424/416: a period average must match the mean of its own dated years

lbm364dl:
*This PR was created by Claude.*

A new detector. **It is the only IIA magnitude test that needs no second opinion.** Every existing one compares a cell to something outside it — another volume (`26_edition_conflicts.py`), another source (`39_cross_source_agreement.py`), the label's own pre-1934 history (`29_era_shift_verdicts.py`), or an absolute threshold. #424 exists *precisely* because its x10 class sits under the 20× floor the collapse and spike screens share.

A period row and the dated rows inside its span are the **same series**, so their relationship is fixed by arithmetic: `1934-1938` must be about the mean of 1934..1938.

## The control is built in, and it is what makes the tails mean anything

```
iia_1925_26   n= 224   median ratio 1.000     (the 1909-1913 average)
iia_1938_39   n= 642   median ratio 1.005     (1928-1932)
iia_1939_45   n= 789   median ratio 1.000     (1934-1938)
                       1,655 comparable pairs
```

The source computes its own averages correctly **as a rule**, so a power-of-ten ratio is a defect rather than an artefact of how the table pairs rows. Arm B pins that median **per volume and bidirectionally** — it is the arm that matters, because the count arms only say *how many* outliers there are. A threshold test has no way to say what normal looks like; this one does.

## Findings

**20 pairs within 15% of 100×, 10 within 15% of a tenth.**

- **18 of the 20 are tobacco (16) or hops (2)**, all in `1928-1932` — #416's inflation arriving through the period rows. That corroborates #416 **at series level**, where its own 500,000 t screen cannot separate a large producer from an inflated one.
- **4 of the 10 low cases are hops *area*** — the second fault already recorded in `iia-hops-x100`, reached here on a slice that entry does not use (it measures the 1933 two-volume overlap; this measures 1934-1938 inside one volume).
- **Two x100 cells fall outside every recorded item scope.**

### `israel / olives / 1934-1938` = 3,075,400 t

Impossible on **four independent comparisons**:

| comparison | value |
|---|---|
| its own dated mean (1934–38) | 30,660 t → **100.3×** |
| its own sibling period rows | 15,751 (1925-29), 12,300 (1928-32) |
| its **paired area row** | 52,000 vs 51,750 ha — **clean**, ratio 1.005 |
| implied yield | **59.1 t/ha** vs 0.59 — real olive yields are 1–3 t/ha |
| Mediterranean total | 3 Mt exceeds it |

That is **#416's exact signature — production ×100, area clean — on an item outside its scope.** The area row being consistent is what turns a large number into an impossible *yield*, so arm F pins both rows and both yields.

Also `indonesia / cotton lint / 1909-1913` at 102×, in `iia_1925_26`, a volume otherwise clean.

## Which side is wrong is deliberately not decided

Both directions occur. For `czech republic / hops / ha` the period figure (11,300) is the historically plausible one and the dated mean (112,800) exceeds world hops area; for `bulgaria / sugar raw centrifugal` the dated mean (16,140) is plausible and the period figure (1,600) is the low one. `verdict` records **the disagreement and its factor, never a culprit.**

## Verification

- **Ten arms (A–J), all verified to fire.**
- **Arm J exists because mutation-testing this gate found a gap in it**: changing `period_value` alone left every other arm quiet, because each read either the value or the ratio and nothing compared them. A derived column can be wrong while its neighbours are right.
- The selftest case relabels one row's volume — moving the control while leaving every ratio, verdict and count **identical** — and fires **arm B alone**.
- `--check` confirms a byte-exact rebuild; `selftest_gates.py` case 12 PASS, and the harness confirms all 90 gates run in CI, are named in the README, and have a case.
- Records a negative: `1925-1929` yields **no** comparable pair, because no series has 3+ dated years in that window.


lbm364dl:
*Comment by Claude.* I mined this detector's two `unexplained_*` buckets (57 + 56 rows) for defects nothing else classifies. **One registered (#619), one hypothesis refuted, and the buckets turn out to hold two different things.**

## Registered: `libya / olives / 1934-1938`

1,610,700 t at **181×** its own dated span, with its **paired area row clean** (61,000 ha vs a dated 60,750, ratio 1.004) giving 26.40 t/ha against a real 1–3, and fao1952 publishing 11,000 t for the same period. Landed in #619 with the item-level control pinned — olives' median era break across 8 labels is **1.195**, so it is two bad cells, not a bad item.

## Refuted: `greece / grapes / 1934-1938`

This looked like the same shape — 4,500 t against a dated mean of 272,960, a ratio of **0.016**, the most extreme entry in `unexplained_dated_higher`. It is not. **The dated series is bimodal:**

```
iia / greece / grapes / tonnes
  1933    1,600      1936   680,100
  1934    3,300      1938   672,500
  1935    5,600      1939   619,900
  1937    3,300      1941/44/45   0
```

Four dated values sit at 1,600–5,600 and three at 620,000–680,000 — a **200× split inside the dated rows themselves**. So the "dated mean" of 272,960 is an average of two incompatible scales and is not a baseline at all. Greece produced on the order of 600,000+ t of grapes in the 1930s, so the *high* mode is the plausible one and the period row agrees with the *low* mode — meaning the period row is not the defect here, it shares one.

There is also no dated area series (area exists only as period rows), so the yield test that settled Libya cannot be run.

**This matters for how the buckets should be read.** `unexplained_dated_higher` conflates two cases: a period row that is wrong, and a period row that is *right* while its dated span is corrupted. The ratio alone cannot tell them apart — only opening the series does. Libya was the first kind, Greece the second.

## What the buckets actually hold

Grapes as a whole is well-calibrated, which is what makes its outliers legible: **37 comparable period rows, median ratio 0.989**, only 3 outside 0.2–5.0. Two of those (morocco 0.113, czech republic 0.133) are already in this table's `disagree_x10_dated_higher` class. Greece is the only one left over, and it is the bimodal case above.

`unexplained_period_higher` is different again — of its 57 rows, **12 are tobacco** in `1928-1932` from `iia_1938_39`, i.e. #416's class with ratios spread 69–26,591× rather than clustered near 100 because the dated baseline is tiny. `mozambique` at 26,591× is 39,000 against a dated mean of **1**. Those are not new defects; they are #416 seen through a degenerate denominator.

So the honest yield of 113 unexplained rows is: one clean new cell, a dozen restatements of #416, and a reminder that a ratio needs its denominator inspected before it means anything.

lbm364dl:
*Comment by Claude, following the Greece case from my previous comment.* **Cross-source settles which mode is wrong, and it is the opposite of what the existing detectors recorded.**

## The evidence

```
iia / greece / grapes / tonnes
  1933   1,600     1936   680,100
  1934   3,300     1938   672,500
  1935   5,600     1939   619,900
  1937   3,300     period 1934-1938   4,500

fao1952 / Greece / grapes   period 1934-1938   450,000 t
juan    / greece / grapes   1911-1932          133,000 - 269,000 t continuously
                            1946-1959          491,000 - 1,233,000 t
```

So Greek grape production ran in the hundreds of thousands of tonnes throughout, and fao1952 puts the **same 1934-1938 period** at 450,000 t.

**The high mode is right and the low mode is a ~100x deficit.** 450,000 / 3,300 = 136x; iia's 680,100 is 1.5x fao's period figure, which is ordinary for a single good year against a five-year mean. Four dated cells (1933, 1934, 1935, 1937) and the `1934-1938` period row are all about two orders of magnitude too small.

## The detectors fired on the wrong cell

```
isolated_spikes.csv    iia / greece / grapes / 1936 = 680,100      <- flagged as a SPIKE
series_collapses.csv   iia / greece / grapes / 1937  interior_collapse
```

`isolated_spikes` recorded **1936 as the anomaly**, because it stands ~200x above its neighbours. On the cross-source evidence 1936 is the **correct** value and its neighbours are the defect. The spike detector identified the right *series* for the wrong *reason*, and its recorded cell is the sound one.

`series_collapses` got 1937 right, as an interior collapse. So the repo currently holds both readings — a spike at 1936 and a collapse at 1937 — describing one alternating series as two anomalies pointing opposite ways.

That is the same hazard as the tobacco rows in `unexplained_period_higher`: **a ratio is only as good as its denominator**, and when a series alternates between two scales, every neighbour-based detector will fire on whichever cell sits on the wrong side of the local comparison.

## Why this one needs no registration from me

Both cells are already in tracked tables, so the class is not invisible — what is wrong is which cell each table names. That is a re-reading of existing records rather than a new defect, and it belongs with whoever owns the spike and collapse screens (#419 / #421) rather than in `data_errors.csv` as a fresh entry. I have not edited either table.

The transferable point is the one worth keeping: **`isolated_spikes` cannot tell a high outlier from a correct value surrounded by low ones.** Cross-source is what separates them, and for Greece it was available — fao1952 and juan both carry the series, and juan carries it continuously across the whole window.

lbm364dl:
*Comment by Claude.* Generalising the Greece case: **how many `isolated_spikes` entries are actually the correct value, with the neighbours as the defect?** Cross-source can answer it, and the answer turns on the period axis.

## The test only works if period rows count as comparators

I ran it three ways. The first two returned **zero adjudicable cases**:

```
comparator = another source, SAME year          0 of 21   (13 have no same-year second source)
comparator = another source, dated +/-3 years   0 of 21   (13 have none in the window)
comparator = above, PLUS period rows whose
             span contains the spike year       2 of 21
```

Both usable comparators are **period rows** (`dated 0, period 1` in each case). My first two attempts filtered with `year.between(...)`, which silently drops rows with a null `year` — the exact trap I documented in #618 today, hit in my own analysis a few hours later.

## Both adjudicated spikes are corroborated, not refuted

```
source  label     item          year   flagged spike   its neighbour   other source
iia     greece    grapes        1936         680,100           5,600        450,000
iia     malawi    cotton lint   1935           3,400             100          3,000
```

In both, another source agrees with the **flagged value** and not with the neighbour it was flagged against. So `isolated_spikes.csv` has named the *sound* cell in 2 of the 21 rows, and the defect is the surrounding low values.

Greece I set out above: fao1952 puts the same `1934-1938` period at 450,000 t and juan carries 133,000–269,000 t continuously through 1911–1932, so 680,100 is ordinary and 1,600–5,600 is a ~100× deficit. Malawi is the same shape — 3,400 t of cotton lint against a neighbour of 100, with another source's period average at 3,000.

## What this does and does not establish

**It does not mean the table is 10% wrong.** 18 of 21 rows cannot be tested at all — 11 have no second source in any form, and 7 I could not map to a polity through `matched_rows` (a label-normalisation failure on my side, not a data fact). The measured statement is narrow: **of the 2 spikes that cross-source can adjudicate, both are the correct value.**

That is a small sample pointing one way, and the direction is the one the detector cannot self-check: a neighbour-ratio screen has no way to tell a high outlier from a correct value surrounded by low ones. It fires either way and records the wrong cell half the time by construction.

**The cheap improvement** is the one this test stumbled into: when a spike has no same-year second source, a period average whose span contains the spike year is often available and is a valid comparator. That took the adjudicable set from 0 to 2 here, and `state/period_volume_provenance.csv` already attributes every iia period row to its volume, so the join is cheap.

lbm364dl:
*Comment by Claude.* Following the two Czechoslovak grapes leads from the scale-slip search. **They are a real defect, they are not a scale slip, and they give the detector a third failure mode.**

## iia's Czech grape PRODUCTION is ~11x deflated from 1932; its AREA is correct

The two sources agree **to the decimal** before the break and diverge after it:

```
year   iia tonnes    juan tonnes        iia ha    juan ha
1930     73,236.8       73,236.8        --        17,500
1931     70,799.8       70,799.8        --        18,800
1932      6,671.4       74,181.4        --        19,100     <- iia falls ~11x
1933      6,800.0       60,500.0        17,298    20,000
1934      4,500.0       55,900.0        21,000    21,000     <- area still identical
1935      8,600.0      102,200.0        24,000    24,000
1936      8,400.0       93,900.0        25,000    25,000
1937     11,100.0      100,600.0        26,000    26,000
```

**73,236.8 and 70,799.8 match to one decimal place in both sources**, so these series share an origin — which makes the post-1932 divergence a defect rather than a difference of scope or definition. And the **area agrees exactly** at 21,000 / 24,000 / 25,000 / 26,000 right through the divergence, so only the production axis moved.

fao1952 corroborates independently: `Czechoslovakia` grapes `1934-1938` = **63,000 t / 19,000 ha**, against iia's 8,100 t. juan's 1934-1938 dated mean is about 88,000 t.

## A second, separate defect on the same series

iia's **area period row** for `1934-1938` reads **2,000 ha** while its own dated areas for those years are 21,000–27,000. So the period row is deflated ~12x on an axis whose dated rows are sound — unrelated to the production problem above.

## Which is the third way the scale-slip signature misfires

My search flagged this pair because at period level iia reads 8,100 t / 2,000 ha and fao reads 63,000 t / 19,000 ha — yields 4.05 and 3.32, close enough to look like one factor applied to both axes. **It is not.** It is a production deflation and an area deflation with independent causes that happen to be of similar size.

So the signature fires on:
1. two different crops with similar yields (19 of 25 — the `#375` wheat/spelt class),
2. a constant-run placeholder on one side (1),
3. **two independent errors of coincidentally similar magnitude** (these grapes cases),

and on a genuine common-factor slip (Malawi, #620). Four causes, one signature. That is weaker than I judged it an hour ago, and I would not gate on it.

## Not registered

The production deflation is well evidenced and worth an entry, but it needs a decision I should not take alone: iia and juan agree exactly for 1930–31, which means one of them is *derived* from the other, and which is upstream determines whether this is an iia defect or a shared one. `20_item_provenance.py` would be the place to settle it. I have recorded the measurement here rather than banking a verdict.

lbm364dl:
*Comment by Claude — correcting my comment above on Czech grapes. I called it "a real defect" and "a production deflation", and said the exact 1930-31 agreement made the divergence "a defect rather than a difference of scope". **It is a difference of scope, it is not a defect, and the repo already records it.**

## The raw extract carries both products side by side

```
raw czechoslovakia, vineyard PRODUCTION, iia_1938_39
  year   vineyards: all      vineyards: grapes, raisins
  1933         60,500.0                     6,800.0
  1934         55,900.0                     4,500.0
  1935        102,200.0                     8,600.0
  1936         93,900.0                     8,400.0
  1937        100,600.0                    11,100.0
```

Layer B's iia `grapes` takes **`vineyards: grapes, raisins`** — the small component. juan's `grapes` matches **`vineyards: all`** to the digit (55,900 / 102,200 / 93,900 / 100,600). Both values are correct for the product each carries. There is no deflation; the two series measure different things.

That also explains the pre-1932 agreement I leaned on: the raw carries a single vineyard product for those years (`vineyards: wine` in `iia_1929_30`), so there was nothing to choose between. The split appears from 1933 in `iia_1938_39`, and layer B picks the smaller side from then on. My "break at 1932" was the point at which the source started offering two products, not the point at which a value went wrong.

## And `item_equivalences.csv` already holds it

```
item=grapes  raw_product=grapes: table  cells=46  series=15  verdict=unresolved
note: Table grapes are a small subset of total vineyard output; this item elsewhere
      carries `vineyards: all`. czech republic reads 8 ha here against 27,000 ha from
      `vineyards: all` in the same series.
```

So the scope mixture is recorded, with a verdict of `unresolved`, and the note even names the Czech case.

## What it actually is

This is **#490's class** — a series publishing a *minor component* where the item name implies the total — appearing on grapes rather than on fertilizer nutrients. #490 measured it for `p`/`n`/`k` and found 8 confirmed series; `item_equivalences.csv` shows the same shape reaching `grapes`, and its `unresolved` verdict is the honest state of it.

I should have checked `item_equivalences.csv` before writing the earlier comment. It is the table that owns "which raw product does this item carry", and I went to `item_provenance.csv` instead, which only answers "which raw *label*".

**The area observation stands and is separate:** iia's `1934-1938` area period row reads 2,000 ha against its own dated areas of 21,000–27,000. That one is not a product-scope question — `item_provenance` attributes the area series to `vineyards: all` with `product_agrees=no` — and it remains unexplained.

lbm364dl:
*Comment by Claude, closing the Czech grapes thread.* **The area anomaly I left "unexplained" is the same component selection — and the way it arises is structurally new.**

## The raw carries both area products too

```
raw czechoslovakia, AREA, 1934-1938 period      raw australia, AREA, 1934-1938 period
  vineyards: all                24,000 ha         vineyards: all                49,000 ha
  vineyards: grapes, raisins     2,000 ha         vineyards: grapes, raisins     3,000 ha

layer-B iia dated areas 1934-1938                 layer-B iia dated areas 1934-1938
  21,000 / 24,000 / 25,000 / 26,000 / 27,000       47,000 / 47,000 / 48,000 / 50,000 / 50,000
  -> these are `vineyards: all`                    -> these are `vineyards: all`
layer-B iia PERIOD row 1934-1938                  layer-B iia PERIOD row 1934-1938
  2,000  -> `vineyards: grapes, raisins`           3,000  -> `vineyards: grapes, raisins`
```

So the area period rows take the minor product while the dated rows take the total — exactly what happens on the production side. That accounts for all four grapes cases in the deflated-area list (`south africa` 0.047, `australia` 0.062, `czech republic` 0.081, `tunisia` 0.082), and they reduce to the `grapes / grapes: table` mixture already recorded in `item_equivalences.csv` with verdict `unresolved`.

**Nothing here is a new defect.** The whole Czech grapes thread collapses into that one known, unresolved scope question.

## But the mechanism is one no check covers

`21_item_product_switches.py` gates a series switching raw product **from year to year**, and #378 established oscillation as the criterion because a one-way switch cannot be told from growth. This is a different axis: **the same layer-B series takes one raw product for its dated rows and a different one for its period rows.**

```
iia / czech republic / grapes / ha
  dated 1934-1938   <- vineyards: all
  period 1934-1938  <- vineyards: grapes, raisins
```

There is no oscillation to detect — each half is internally consistent — and the two halves are never compared to each other by any product-level check. It is only visible because a period average and its own dated years must agree, which is what this detector measures.

That makes it the fourth thing the dated/period axis has turned up today, after #414's zero count, #416's row scope and #443's phantom duplicate year. Worth stating as a general shape: **a layer-B series is not guaranteed to draw its dated and period rows from the same raw product**, and nothing currently checks that it does.

## Where the thread ends

- The production "deflation" I claimed: **retracted** — correct data, different products.
- The area anomaly: **explained** — same cause, same `unresolved` entry.
- The genuinely new part: **the dated/period product split**, which is a gap in `21_item_product_switches.py`'s coverage rather than a defect in any cell.

The 5 extreme `1928-1932` cases in the same list are a different story again — `india / sesame seed` at 1,000 ha against a dated mean of 2,363,237, and its siblings, are #366's constant-run placeholders sitting in the period row.

